PR #8, opened 2021-11-02
Add daily nwis data to project pipeline

lekoenig:
This PR is to add daily mean SpC data from NWIS to the project pipeline. First, to make the _targets.R file more interpretable as the pipeline grows, I've refactored the pipeline to include different makefiles that correspond to a single phase (e.g. "1_fetch.R","2_process.R"). The phase makefiles are then sourced and combined within _targets.R.

In addition to pipeline organization, the code now:

- identifies NWIS sites within the DRB for which specific conductance data is available
- downloads daily (mean) SpC values for the available period of record for sites with daily data
- munges individual site data and outputs a uniform format
- includes the NWIS sites in a leaflet map within the salinity data report to show spatial coverage and sample density of SpC data in the DRB

I'd welcome any feedback you have but in particular, I'd be happy to discuss the format of the input data we'll want for the modeling. Here I've assumed we're most interested in daily mean specific conductance, but perhaps the 15-min data from these sites would also be useful. Thanks!


jds485:
I haven't looked at code yet, but just some thoughts before viewing:

Did you cross check these data with the harmonized WQP dataset? If not, that can be a separate PR.

I am also interested in the 15-min data for use in a sub-daily prediction model. Up to you if it makes more sense to have that in this PR or in a separate PR.

lekoenig:
> Did you cross check these data with the harmonized WQP dataset? If not, that can be a separate PR.

Yep! In general, the WQP focuses on discrete samples rather than continuous data, but to make sure we're not generating redundant datasets I cross-checked the downloaded data against the harmonized WQP dataset for a subset of sites. The data downloaded in this PR are not represented in the harmonized WQP data. The target `p1_SpC_sites` currently returns all sites with specific conductance data regardless of the "service code." However, we only download data from nwis for sites tagged as [daily data](https://help.waterdata.usgs.gov/faq/water-quality/water-quality-daily-data) (service = "dv"). This download step takes place within the `p1_nwis_daily_SpC_data_ls` target. Similarly, the 15-min data is not included in the harmonized WQP dataset. 

> I am also interested in the 15-min data for use in a sub-daily prediction model. Up to you if it makes more sense to have that in this PR or in a separate PR.

OK, great. I can add that download step and commit to this PR.

jds485:
For the 15-min data, Joel Blomquist suggesting aggregating to hourly to reduce noise in the timeseries. Hourly is probably also better for working with hourly weather data

lekoenig:
> For the 15-min data, Joel Blomquist suggesting aggregating to hourly to reduce noise in the timeseries. Hourly is probably also better for working with weather data

Yeah, this aggregation probably makes sense for a few reasons, including that any sub-daily model would probably not target 15-min predictions. I think it makes sense to go ahead and do this sort of processing on the instantaneous data, so I'm going to revert back to having this PR focus on the daily values. I'll open a separate PR for the instantaneous data.

lekoenig:
@jds485 For this PR review, I'm most interested in discussing the format of the specific conductance observations that we want to target for modeling. For example, for a daily salinity model, I've assumed we're most interested in the daily means as opposed to min/max/etc. Let me know if you have other thoughts on that.  Thanks!

_targets.R
```diff
@@ -1,9 +1,11 @@
 library(targets)
 
 options(tidyverse.quiet = TRUE)
-tar_option_set(packages = c("tidyverse", "lubridate","rmarkdown","knitr","leaflet","sf")) 
+tar_option_set(packages = c("tidyverse", "lubridate","rmarkdown","dataRetrieval","knitr","leaflet","sf")) 
 
-source("./2_process/src/2_process.R")
+source("1_fetch.R")
+source("2_process.R")
+source("3_visualize.R")
```
jds485, line 8: I'm not 100% sure, but I think USGS practice is to place these in their respective src folders instead of the main directory. I kind of like seeing them in the main repo for quick access, but it's probably not as clean once there are many steps in the pipeline.

2_process.R
```diff
@@ -0,0 +1,19 @@
...
+    filter_wqp_salinity_data(p1_wqp_data,major_ion_names,select_wqp_vars,omit_wqp_events)
+  ),
+  # Subset and save discrete SpC data from harmonized WQP
+  tar_target(
+    p2_wqp_spC_csv,
+    subset_wqp_spC_data(p2_filtered_wqp_data,fileout="2_process/out/DRB_WQdata_SpC_data.csv")),
+  # Clean and save NWIS daily SpC data 
+  tar_target(
+    p2_nwis_SpC_daily_csv,
+    combine_daily_mean_SpC_data(p1_nwis_SpC_daily_data,fileout="2_process/out/DRB_daily_SpC_data.csv")
```
jds485, line 17: missing `format = 'file'` for these 2 targets

3_visualize.R
```diff
@@ -0,0 +1,6 @@
+
+p3_targets_list <- list(
+  # Render data summary report
+  tarchetypes::tar_render(p3_wqp_spC_report, "3_visualize/src/report-wqp-salinity-data.Rmd",output_dir = "3_visualize/out")
```
jds485, line 4: It might help to note in a comment that tar_render() returns a `format = 'file'` target

_targets.R
lekoenig: I think these are needed though so that the _targets.R file knows where to access the lists of targets for the individual phases.

2_process.R
lekoenig: Great catch, thanks! I'm guessing the pipeline only builds because of the way I use tar_read in the rmarkdown report. I'll make this change.

_targets.R
```diff
@@ -22,23 +24,19 @@ major_ion_names = c("Chloride","Sodium")
...
+drb_huc8s <- c("02040101","02040102","02040103","02040104","02040105","02040106",
+               "02040201","02040202","02040203","02040204","02040205","02040206","02040207",
+               "02040301","02040302")
+
+# Define USGS site types for which to download specific conductance data (include "Stream","Stream:Canal", and "Spring" sites)
+site_tp_select <- c("ST","ST-CA","SP") 
+
+
+# Return the complete list of targets
+c(p1_targets_list, p2_targets_list, p3_targets_list)
```
jds485, line 40: I didn't notice this when I went through targets trainings, but these lists are unnamed. So, I'm not sure how to load the targets to inspect them. Did you find a way to do that? And should we name these lists with the target names?

_targets.R
jds485: Yes, they're definitely needed! I'm just suggesting to move them to the src directories

2_process.R
```diff
@@ -0,0 +1,19 @@
...
+p2_targets_list <- list(
+  # Filter harmonized WQP data for salinity data
+  tar_target(
+    p2_filtered_wqp_data,
+    filter_wqp_salinity_data(p1_wqp_data,major_ion_names,select_wqp_vars,omit_wqp_events)
+  ),
+  # Subset and save discrete SpC data from harmonized WQP
+  tar_target(
+    p2_wqp_spC_csv,
+    subset_wqp_spC_data(p2_filtered_wqp_data,fileout="2_process/out/DRB_WQdata_SpC_data.csv")),
```
jds485, line 13: Could remove NA values. They are all listed as equipment maintenance or *** provisional codes

jds485: Sorry, placed this comment on the wrong dataset. Supposed to be the DRB_daily_SpC_data.csv

jds485:
> For example, for a daily salinity model, I've assumed we're most interested in the daily means as opposed to min/max/etc

I think max for sure is of interest. We could probably generate more robust statistics to use, like 5th percentile (estimated as 5th largest daily measurement when there are 15 minute samples). That's thinking back to Joel's point about aggregating the 15 min data. But as far as built-in NWIS codes, max would be helpful to have.

jds485 (approved):
Code looks good and runs for me! A few comments. We can move to discussion if that's better.

_targets.R
lekoenig: Ah, I see! I first thought you were suggesting moving these calls to source as opposed to just changing the path of `1_fetch.R`. I kind of like keeping the phase targets files in the main directory to separate these lists of targets from the more specific functions in the respective src folders, but I know what you mean about the main directory getting cluttered.

After taking a quick look back at the formatting used in the targets training as well as another USGS-R repo (delaware-model-prep), I decided to keep these files in the main project directory for now, but we can revisit as we add other steps to the pipeline.

_targets.R
lekoenig: Yeah, I adopted the naming conventions from the targets trainings for these lists, but they could be made more informative, for example, by describing the phase (e.g. `p1_observations_targets_list`). I've kept the phase number naming convention for now, but let's revisit once we start adding modeling phases to the pipeline. 

I think these lists are more organizational than functional from a targets perspective. The lists keep the main `_targets.R` file cleaner by separating the targets out by phase, but to my knowledge you still need to call individual targets in order to inspect them using `tar_load` or `tar_read`. So I can't call `tar_load(p1_targets_list)` but I can call `tar_load(p1_wqp_data)` within the `p1_targets_list`.

2_process.R
lekoenig: Good point! This relates to a question I had for the continuous data too - will we want padded time series, or just the available data? Does it make a difference from a modeling perspective?

lekoenig: Questions about time series formatting migrated to discussion (#15). I've gone ahead and removed NA values from the daily data summaries since they all reference maintenance or other equipment codes.

lekoenig:
> I think max for sure is of interest. 

I've added max SC to the daily outputs, too. In doing this, I ended up changing how I was mungeing the column names. A few sites have unusual or unexpected column names, which I now handle on a site-specific basis in the `munge_daily_mean_SpC_cols` function (`munge_nwis_SpC_data.R`). For example, a new column name might be adopted when an instrument gets moved or a new instrument is deployed. Only using our expected `SpecCond` column means potentially missing out on this new data. So for these few sites, I merge the old and new time series 1) after checking the site documentation that both measures can reasonably be assumed to represent river channel SC, and 2) after inspecting the periods of overlap for agreement in SC values between the two columns. In many ways, it would be nice to avoid site-specific data mungeing, but I've seen this done in other USGS-R repos, and this new code chunk ensures we're not leaving information on the table.

@jds485 Other than adding maximum values, not much has changed here, but let me know if you want to re-review prior to me merging this PR.

jds485:
Site-specific data processing is fine for the case you're describing. I've had column name issues with USGS data before. I'll review those changes and make sure the code runs for me before it's pushed to the USGS repo.

jds485:
There are 28 samples for which the max is less than the mean. Do you think we should drop these?

1_fetch.R
```diff
@@ -0,0 +1,26 @@
...
+
+p1_targets_list <- list(
+  # Load harmonized WQP data product
+  tar_target(
+    p1_wqp_data,
+    readRDS(file = "1_fetch/in/DRB.WQdata.rds")),
+  tar_target(
+    p1_nwis_SpC_sites,
+    {
+      dummy <- '2021-11-12'
```
jds485, line 11: I think it will be useful to have this date set within the main targets script. One way to do that: place `dummy_date <- '2021-11-12'` into the main targets script, and change this line to `dummy <- dummy_date`

jds485: Oh, just thinking about this suggestion, I think that will force a new download to today's date instead of 11-12

lekoenig: Good suggestion, that placement makes more sense! I ran into a situation where I updated the data download step to include max SC, but a site within `p1_nwis_SpC_sites` had been taken offline in the meantime (a site with 2 recorded observations, so perhaps it was a preliminary/test run of some kind). So the dummy variable is used to force a rebuild of the nwis site list if `dummy_date` is changed from `2021-11-12`.

lekoenig:
> There are 28 samples for which the max is less than the mean. Do you think we should drop these?

@jds485 Yes, I think we should go ahead and drop these few samples for which daily max SC < daily mean SC. These 28 instances seem to be sporadic across a number of sites and are all older observations anyway (prior to 1990). I've implemented this filtering step in "2_process/src/munge_nwis_SpC_data.R"

Issue #13, opened 2021-11-08
Add daily maximum to processed daily/continuous SpC data


lekoenig:
> For example, for a daily salinity model, I've assumed we're most interested in the daily means as opposed to min/max/etc

I think max for sure is of interest. We could probably generate more robust statistics to use, like 5th percentile (estimated as 5th largest daily measurement when there are 15 minute samples). That's thinking back to Joel's point about aggregating the 15 min data. But as far as built-in NWIS codes, max would be helpful to have.

_Originally posted by @jds485 in https://github.com/USGS-R/drb-inland-salinity-ml/issues/8#issuecomment-962233263_

Issue #11, opened 2021-11-05
Aggregate 15 min conductance data to hourly outputs


lekoenig:
> For the 15-min data, Joel Blomquist suggesting aggregating to hourly to reduce noise in the timeseries. Hourly is probably also better for working with weather data

Yeah, this aggregation probably makes sense for a few reasons, including that any sub-daily model would probably not target 15-min predictions. I think it makes sense to go ahead and do this sort of processing on the instantaneous data, so I'm going to revert back to having this PR focus on the daily values. I'll open a separate PR for the instantaneous data.

_Originally posted by @lekoenig in https://github.com/USGS-R/drb-inland-salinity-ml/issues/8#issuecomment-960692681_

Issue #9, opened 2021-11-02
Check that we are pulling all available NGWOS data


lekoenig:
Check that our SpC data download from NWIS daily sites (service = "dv") includes all available NGWOS sensor sites. 

@jds485 & @aappling-usgs, are you aware of a published list of NGWOS stations in the DRB? I found a list of 124 sites within the IIDD --> Project - Modeling Streams docs, but I'm not sure if it's still up to date.

jds485:
No, I don't know of a list with those stations. I'd ask Joel Blomquist.

jds485:
@lekoenig will this issue be resolved with the PR #19, or are you waiting on further info about NGWOS sites? If 19 will close the issue, you can link the pull request to this (panel on right).

lekoenig:
This issue will be closed with PR#19. I don't incorporate this check as a target within the pipeline, but we are capturing all 68 NGWOS SC sites within the DRB. I just linked this issue to the pull request, thanks for that suggestion!

PR #19, opened 2021-11-30
Add instantaneous nwis data to project pipeline

lekoenig:
This PR adds instantaneous SC data from NWIS to the project pipeline. I've made some edits to the download of daily summary data - namely to take advantage of dynamic branching across sites within the targets pipeline - but I've also added max SC to the daily summaries download step for sites where it's available (issue #13). 

For the instantaneous data, the following steps are included here:

- download sub-daily SC values for the available period of record for sites with instantaneous data
- do initial cleaning on individual site data and output a uniform format
- aggregate sub-hourly SC values to hourly outputs (issue #11)
- create plots of individual time series (_are these helpful to others, or do they clutter your project directory too much?_) 

The sites included here include all of the non-tidal NGWOS sensor sites within the DRB (n = 67 right now; issue #9). The next step is to create model-ready SC outputs across all of our sites, but I wanted to get these data in the pipeline separate from some of the other mungeing steps I've been working on to keep the steps/diffs more manageable for PR reviews.

hourly_data_01467000
Rplot01


3_visualize/src/plot_inst_data.R
```diff
@@ -0,0 +1,38 @@
...
+  
+  message(sprintf('  Plotting instantaneous data for %s', site_data$site_no[1]))
+  
+  # Check that site_data contains required columns
+  req_cols <- c("Value_Inst_hourly","Value_Inst_cd","dateTime","site_no")
+  flag_cols <- req_cols[which(req_cols %in% names(site_data)=="FALSE")]
+  if(length(flag_cols)>0) stop("Input data is missing one or more required columns: Value_Inst_hourly, Value_Inst_cd, dateTime, site_no")
+  
+  # Define y-axis label
+  ylabel <- case_when(site_data$Parameter[1] %in% c("SpecCond","00095") ~ expression(Hourly~mean~SC~at~'25'~deg~C~(mu~S/cm)),
```
jds485, line 21: `expression(paste())` can help for these special characters. `expression(paste0('Hourly mean SC at 25 ', degree, 'C (', mu, 'S/cm)'))` Check my commas and quotation marks before using

lekoenig: Thanks, I always like to hear different ways people carry out various tasks in R. I'll try this out as a way to blend strings with special characters.

2_process/src/munge_nwis_SpC_data.R
```diff
@@ -1,72 +0,0 @@
-munge_daily_mean_SpC_cols <- function(x){
```
jds485, line 1: Did you use `git mv` to rename a file? Git will track that as a rename instead of a deletion, and I think GitHub would also track this as a rename. Not important to change, just a tip for future

_targets.R
```diff
@@ -24,19 +26,29 @@ major_ion_names = c("Chloride","Sodium")
 # Define hydrologic event types in harmonized WQP data to exclude
 omit_wqp_events <- c("Spill","Volcanic action")
 
-# Define USGS specific conductance parameter codes
-SpC_pcodes <- c("00095","90095","00094","90096") 
+# Define all USGS specific conductance parameter codes
+pcodes_select <- c("00095","90095","00094","90096")
```
msleckman, line 30: Suggestion: Since most of these variables are specific to either `1_fetch.R`, `2_process.R`, it would be easier to have them defined within the phase target scripts rather than in the overarching `_targets.R` if that is possible.

.gitignore
```diff
@@ -4,6 +4,9 @@ _targets
 #All model output
 */out/*
 
+#Other local files
+archive/*
```
jds485, line 8: What is in the archive? Just curious. I guess this must be your local files because it's not showing up when running the pipeline

_targets.R
```diff
@@ -24,19 +26,29 @@ major_ion_names = c("Chloride","Sodium")
...
 omit_wqp_events <- c("Spill","Volcanic action")
 
-# Define USGS specific conductance parameter codes
-SpC_pcodes <- c("00095","90095","00094","90096") 
+# Define all USGS specific conductance parameter codes
+pcodes_select <- c("00095","90095","00094","90096") 
 
-# Define minor HUCs (hydrologic unit codes) that make up the DRB (HUC04 = 0204)to use in calls to dataRetrieval functions
+# Define USGS specific conductance parameter code to use for continuous data
+parameter <- "00095"
```
jds485, line 33: related to comment above, are there other codes for instantaneous data?

1_fetch/src/get_daily_nwis_data.R
```diff
@@ -0,0 +1,63 @@
...
+  
+  message(sprintf('Retrieving daily data for %s', site_info$site_no))
+
+  # Download daily data
+  site_data <- dataRetrieval::readNWISdv(
+    siteNumbers = site_info$site_no,parameterCd=parameter,statCd=stat_cd_select,startDate = "",endDate = "") %>%
+    dataRetrieval::renameNWISColumns(p00300="Value",p00095="Value")
+  
+  # Munge column names for some sites with different or unusually-named value columns
+  if(length(grep("Value_cd",names(site_data)))>1){
```
jds485, line 21: If this gets to be very long, it could become a separate function that passes in the dataframe (or only the problem sites) and returns the processed data

1_fetch/src/get_daily_nwis_data.R
```diff
@@ -0,0 +1,63 @@
...
+  
+  # Filter daily data
+  site_data_out <- site_data %>%
+           # omit rows with no data
+    filter(!(is.na(Value) & is.na(Value_Max)),
+           # omit rows where daily mean > daily max
+           (is.na(Value)|is.na(Value_Max)|(!Value > Value_Max)),
+           # omit rows with undesired data quality codes
+           !(Value_cd %in% c("P Eqp","P Mnt")),
+           !(Value_Max_cd %in% c("P Eqp","P Mnt"))) %>%
```
jds485, line 58: I have seen records that omit spaces, use lowercase, etc. for these codes (I haven't checked if that's the case for our project sites). Maybe grepping for "eqp" and "mnt" is better to catch these oddities (with `ignore.case = TRUE`).

1_fetch/src/get_daily_nwis_data.R
```diff
@@ -0,0 +1,63 @@
+get_daily_nwis_data <- function(site_info,parameter,stat_cd_select) {
```
jds485, line 1: eventually, we'll probably want the start and end date as arguments here. Comment may apply to other functions, too

_targets.R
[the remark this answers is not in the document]
lekoenig: Perhaps I shouldn't say "all" on line 29 here. I originally went through and found all of the SC pcodes but ultimately decided to omit the more extraneous ones and narrow down to this target list that represent both in situ and lab-measured SC values. No DRB sites get returned from any of these other pcodes so I am going to leave `pcodes_select` as is. However, I'll make a note of these other pcodes in the markdown doc since I've been tracking some data processing decisions there.

.gitignore
lekoenig: Yep, just local scripts and notes that I didn't want git to track

_targets.R
lekoenig: I can definitely see how keeping the variables within their respective phase files would be easier to read. I think we initially even started that way but at one point we decided to keep all of the defined variables within the `_targets.R` file so that a user would see them all in one place, as opposed to having to check the individual phase files. I'll suggest we leave as is for now, but we can revisit this sort of formatting if our `_targets.R` file starts to get too unwieldy!

_targets.R
```diff
@@ -24,19 +26,29 @@ major_ion_names = c("Chloride","Sodium")
...
+
+# Define minor HUCs (hydrologic unit codes) that make up the DRB to use in calls to dataRetrieval functions
+# Lower Delaware: 0204 subregion (for now, exclude New Jersey Coastal (https://water.usgs.gov/GIS/huc_name.html)
```
jds485, line 36: did you add this note to the markdown? Are the NJ coastal HUCs the 2 deleted?

_targets.R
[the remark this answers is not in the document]
jds485: Sounds good, thanks!

_targets.R
```diff
@@ -9,9 +9,11 @@ source("3_visualize.R")
 
 dir.create("2_process/out/", showWarnings = FALSE)
 dir.create("3_visualize/out/", showWarnings = FALSE)
+dir.create("3_visualize/out/daily_timeseries_png/",showWarnings = FALSE)
+dir.create("3_visualize/out/hourly_timeseries_png/",showWarnings = FALSE)
```
jds485, line 13: Responding to your question in the PR, I think it's fine to write all of these. I like having them as a quick reference.

_targets.R
lekoenig: My understanding is that "00095" is the code used for in situ SC monitoring. The table below is from the USGS [report](https://nc.water.usgs.gov/reports/tm1D3/pdf/TM1D3.pdf) "Guidelines and standard procedures for continuous water-quality monitors." I also did a query within the DRB and "00095" is the only code returned for instantaneous ("uv") or daily summary ("dv") sites. By the way, did you know that pcodes are getting phased out as part of NWIS/Aquarius modernization? 

Wagner2006_Table23

_targets.R
lekoenig: Sounds good, thanks! They're meant to be a quick reference but we could obviously change the formatting if anyone's opposed to ggplot defaults : )

1_fetch/src/get_daily_nwis_data.R
jds485: I noticed a filter for these in other scripts above, too, so comment also applies wherever they appear. Do you know a way to search the whole repo for text in R? Looks like it is possible within GitHub

_targets.R
lekoenig: I originally included the two NJ Coastal HUCs because they're within the Delaware subregion but have since omitted them - the DRB harmonized WQP dataset does not include those HUCs and neither does the PRMS dataset from the DRB temperature work. Let me know if you think otherwise. Either way, I'll add a note to the markdown doc.

2_process/src/filter_wqp_data.R
```diff
@@ -1,9 +1,19 @@
-filter_wqp_salinity_data <- function(data,major_ion_names,select_wqp_vars,omit_wqp_events,exclude_tidal=TRUE){
-  
-  # major_ion_names are values to keep from the param column of the harmonized wqp data
-  # select_wqp_vars are columns of data to keep from the harmonized wqp data
-  # omit_wqp_events are values to omit from HydrologicEvent column of the harmonized wqp data
-  # exclude_tidal logical, defaults to TRUE. If TRUE, rows containing "tidal" within the MonitoringLocationTypeName column of the harmonized wqp data are omitted
+filter_wqp_salinity_data <- function(data,major_ion_names,wqp_vars_select,omit_wqp_events,exclude_tidal=TRUE){
+  #' 
+  #' @description Function to filter the DRB multisource surface-water-quality dataset for desired parameters, units, and variables
```
jds485, line 3: Thanks for adding in all of the function descriptions!

_targets.R
jds485: Okay, thanks! Great reasons to omit them, then

_targets.R
jds485: Thanks! I did not know that pcodes are going away. Should we be adapting to the new format for this work? And what is the new format? We can discuss outside of this PR if that makes more sense

lekoenig: I learned about it from the DRB data harmonization group - I think the plan is to transition to "observed properties and units" with a rollout in FY21/22. That group didn't seem to know enough about it to change their data workflows and I'm guessing current systems will be supported for awhile. So I suggest we shouldn't worry about altering our workflows at this stage but it's good to know that it's (maybe) coming...

1_fetch/src/get_daily_nwis_data.R
lekoenig: OK, yeah that probably makes sense. I will leave as is for now but can revisit if needed - perhaps I'm too optimistic that the number of 'unique' sites won't grow too much!

1_fetch/src/get_daily_nwis_data.R
lekoenig: Great idea, I'll reformat this to use `grep()` instead. I haven't seen those other variations yet, but I definitely believe you that they're out there! The `grep()` approach will be more robust in any case.

lekoenig: You can search in Rstudio, at least: Ctrl+Shift+F opens a "Find in Files" window that you can use to search for text across files in an R project.

2_process/src/munge_nwis_SpC_data.R
lekoenig: I actually did delete this file `munge_nwis_SpC_data.R` because it was extraneous after I incorporated dynamic branching in the phase file and column name cleaning in the `get_daily_nwis_data.R` file. That's good to know about `git mv`!

1_fetch/src/get_daily_nwis_data.R
lekoenig: Thanks, I agree! For example, specifying an end date would be useful for ensuring consistent builds across users. I've added `start_date` and `end_date` as arguments to `get_daily_nwis_data` and `get_inst_nwis_data`. What I'm thinking is that I'll keep "" as defaults for those arguments (this is what dataRetrieval does), but within our `1_fetch.R` phase file, specifying `dummy_date` as the `end_date` within the target builds for the daily and instantaneous data. That way `dummy_date` is useful for us as an indicator of last build/data download. How does that sound?

_targets.R
jzwart: FYI more details [in this Teams conversation ](https://teams.microsoft.com/l/message/19:e320b241dd91479fba9b83c5bf9f6d99@thread.skype/1638537022782?tenantId=0693b5ba-4b18-4d7b-9341-f32f400a5494&groupId=2651eb2e-8cf1-4caa-a084-0a11facc1d36&parentMessageId=1638537022782&teamName=GS-WMA%20IIDD%20Staff&channelName=General&createdTime=1638537022782)on this transition

lekoenig: Thanks, @jzwart! So it sounds like discrete data ("qw") pulls will be impacted first, but that there will be a ~2 year transition period from pcodes to observed properties (OPs). I assume this plan will also apply to unit values when the time comes.

jzwart: Yes that's my understanding.

1_fetch/src/get_daily_nwis_data.R
jds485: I think that's a great idea

3_visualize/src/report-wqp-salinity-data.Rmd
```diff
@@ -10,26 +10,23 @@ output: html_document
```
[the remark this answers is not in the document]
lekoenig: I'd definitely be open to that - I initially created the markdown doc as a way to show some draft figures and document some notes. My 2 cents is that most of the code within the markdown doc would represent such a small target that it's not worth porting over to a separate function (e.g. code that tallies censored/preliminary data within the harmonized WQP dataset). 

I think it's worth moving some code into functions for figures, though. I'll suggest that I start by moving the leaflet map code into a separate function. Your comment here made me rethink the location of the code that cleans up the different crs within the site list. I've since moved that code from the markdown doc to `2_process/create_site_list.R` instead, which will hopefully make downstream tasks like snapping site locations to PRMS flowlines easier.

jds485:
As discussed, check sites with very large SC values as possibly tidal. Map of these sites with SC > ~10000 could be helpful to diagnose.

jds485:
I just flipped through output timeseries. 01481500 seems like a possible moved gauge/site over time (pre-1980 different than the rest of the timeseries and possibly tidal). 01482695 is also very high.

jds485 (changes requested):
Looks good and the code runs for me. Remaining edit request is to screen sites that are likely tidal

lekoenig:
> I just flipped through output timeseries. 01481500 seems like a possible moved gauge/site over time (pre-1980 different than the rest of the timeseries and possibly tidal). 01482695 is also very high.

I agree that we should omit sites [01484272](https://waterdata.usgs.gov/md/nwis/uv/?site_no=01484272) and [01482695](https://waterdata.usgs.gov/de/nwis/nwismap/?site_no=01482695&agency_cd=USGS). I don't show marker labels on the snapshot map below, but these sites are the two red markers that have mean SC > 6000 uS/cm and are likely estuary-influenced. 01482695 is the "C and D Canal," which connects the Chesapeake Bay with Delaware Bay. 

Another site with some _very_ high SC records is [01467081](https://waterdata.usgs.gov/nj/nwis/uv/?site_no=01467081&agency_cd=USGS). This site doesn't appear to be tidal and is adjacent to a highway. The highest SC records come from Jan/Feb 2021 when there were a couple winter storms. However, I did notice that the remarks for this site mention a sewage plant upstream. We can see some evidence of diel fluctuations in the SC time series here, presumably due to discharge fluctuations. I plan to leave this site for now, but wanted to make a note of it following our discussion of high-SC sites.  

high_sc_sites

And I see what you mean about [01481500](https://waterdata.usgs.gov/usa/nwis/uv?01481500). Daily SC values are so much higher during 1968-1970 compared to more recent records. The site remarks do mention a minor(?) site relocation as well as flow regulation by an upstream reservoir post-1973. I suppose we could omit this site from the daily data records, however if the time period for our project work is set by met data availability (~1980), perhaps these data would be omitted anyway? Alternatively, we could set the startDate as "1979-10-01" for all record requests. @jds485 do you have any other thoughts on that?

daily_data_01481500

jds485:
Thanks for looking into this! Those are very high SC records for 01467081! I'd assume they are storm induced rather than wastewater treatment induced. The NHD attribute "density of NPDES sites" should account for wastewater plants, but I remember Salme noting that the contribution of wastewater plants to salinity is spatially variable. If the diel signals seem to have large amplitudes, we can revisit a different wastewater approach.

If river regulation were the cause for 01481500 being so different, that would be sort of amazing. From a map of the watershed, it looks like only a small portion of the drainage basin is regulated, and that portion is nearly undeveloped, so I'm doubtful regulation would be the cause. 

I think we can safely set the start date to 1979-10-01 due to the met data availability.

jds485 (approved):
Looks good! You can merge when ready

PR #26, opened 2021-12-10
[#24] simple summaries of inst data

### LINKED ISSUE #24: aggregate instantaneous data to daily
We need to aggregate the instantaneous data to daily summaries

jsadler2:
Daily summaries (min/mean/max) of instantaneous data. 

closes #24

### DIFF
--- a/2_process.R
+++ b/2_process.R
@@ -38,12 +38,12 @@ p2_targets_list <- list(
      {
        sites_tbl <- read_csv(p2_site_list_csv)
        get_site_flowlines(p1_reaches_sf, sites_tbl, sites_crs = 4269, max_matches = 1, search_radius = 0.1)
-     })
+     }),
   
   # Aggregate instantaneous DO data to daily min/mean/max
   tar_target(
     p2_inst_data_daily,
-    aggregate_data_to_daily(p1_inst_data,p1_daily_data)),
+    aggregate_data_to_daily(p1_inst_data,p1_daily_data,"America/New_York")),
   
   # Combine 1) daily DO data and 2) instantaneous DO data that has been aggregated to daily 
   tar_target(
--- a/2_process.R
+++ b/2_process.R
@@ -40,7 +40,7 @@ p2_targets_list <- list(
        get_site_flowlines(p1_reaches_sf, sites_tbl, sites_crs = 4269, max_matches = 1, search_radius = 0.1)
      }),
   
-  # Aggregate instantaneous DO data to daily min/mean/max
+  # Aggregate instantaneous SC data to daily min/mean/max
   tar_target(
     p2_inst_data_daily,
     aggregate_data_to_daily(p1_inst_data,p1_daily_data, min_daily_coverage=0.5, output_tz="America/New_York")),
--- a/2_process.R
+++ b/2_process.R
@@ -39,4 +39,14 @@ p2_targets_list <- list(
        sites_tbl <- read_csv(p2_site_list_csv)
        get_site_flowlines(p1_reaches_sf, sites_tbl, sites_crs = 4269, max_matches = 1, search_radius = 0.1)
      })
+  
+  # Aggregate instantaneous DO data to daily min/mean/max
+  tar_target(
+    p2_inst_data_daily,
+    aggregate_data_to_daily(p1_inst_data,p1_daily_data)),
+  
+  # Combine 1) daily DO data and 2) instantaneous DO data that has been aggregated to daily 
+  tar_target(
+    p2_daily_combined,
+    bind_rows(p1_daily_data, p2_inst_data_daily))
 )
--- a/2_process.R
+++ b/2_process.R
@@ -43,7 +43,7 @@ p2_targets_list <- list(
   # Aggregate instantaneous DO data to daily min/mean/max
   tar_target(
     p2_inst_data_daily,
-    aggregate_data_to_daily(p1_inst_data,p1_daily_data,"America/New_York")),
+    aggregate_data_to_daily(p1_inst_data,p1_daily_data, min_daily_coverage=0.5, output_tz="America/New_York")),
   
   # Combine 1) daily DO data and 2) instantaneous DO data that has been aggregated to daily 
   tar_target(
--- a/2_process/src/munge_inst_timeseries.R
+++ b/2_process/src/munge_inst_timeseries.R
@@ -78,9 +78,9 @@ aggregate_data_to_daily <- function(inst_data, daily_data, output_tz){
     mutate(dateTime = with_tz(dateTime,tzone=output_tz)) %>%
     mutate(Date = date(dateTime)) %>%
     group_by(Date, site_no) %>%
-    summarise(Value = mean(Value_Inst), 
-              Value_Min = min(Value_Inst), 
-              Value_Max = max(Value_Inst), 
+    summarise(Value = mean(Value_Inst, na.rm=TRUE), 
+              Value_Min = min(Value_Inst, na.rm=TRUE), 
+              Value_Max = max(Value_Inst,na.rm=TRUE), 
               na_count=sum(is.na(Value_Inst)), 
               value_count=sum(!is.na(Value_Inst))) %>%
     mutate(percent_coverage=value_count/(value_count + na_count)) %>%
--- a/2_process/src/munge_inst_timeseries.R
+++ b/2_process/src/munge_inst_timeseries.R
@@ -75,9 +75,9 @@ aggregate_data_to_daily <- function(inst_data, daily_data, output_tz){
   
   daily_values <- inst_data %>%
     filter(site_no %in% only_inst_data) %>%
-    mutate(dateTime = with_tz(dateTime,tzone=output_tz)) %>%
-    mutate(Date = date(dateTime)) %>%
-    group_by(Date, site_no) %>%
+    mutate(dateTime_local = lubridate::with_tz(dateTime,tzone=output_tz),
+           Date = lubridate::date(dateTime_local)) %>%
+    group_by(Date, site_no, .groups = "keep") %>%
     summarise(Value = mean(Value_Inst, na.rm=TRUE), 
               Value_Min = min(Value_Inst, na.rm=TRUE), 
               Value_Max = max(Value_Inst,na.rm=TRUE), 
--- a/2_process/src/munge_inst_timeseries.R
+++ b/2_process/src/munge_inst_timeseries.R
@@ -55,3 +55,26 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
 
 }
 
+aggregate_data_to_daily <- function(inst_data, daily_data){
+  #' 
+  #' @description Function to aggregate instantaneous NWIS data collected at sub-hourly (e.g. 15/5/30 min) intervals to hourly min/mean/maxs
+  #'
+  #' @param inst_data a data frame containing the downloaded time series for NWIS instantaneous site. 
+  #' inst_data must include the following columns: c("Value_Inst","dateTime","site_no")
+  #' @param daily_data a data frame the downloaded daily time series of DO data. This is used so that if a site is already in 
+  #'  the daily sites, we won't do the aggregating here
+  #'
+  #' @value A data frame containing daily min, mean, and max values the parameter of interest
+  
+  
+  only_inst_data = setdiff(inst_data$site_no, daily_data$site_no)
+  
+  daily_values <- inst_data %>%
+    filter(site_no %in% only_inst_data) %>%
+    mutate(Date = as.Date(dateTime, format="%Y-%m-%d")) %>%
+    group_by(Date, site_no) %>%
+    summarise(Value = mean(Value_Inst), Value_Min = min(Value_Inst), Value_Max = max(Value_Inst)) %>%
+    na.omit()
+  
+  return(daily_values)
+}
--- a/2_process/src/munge_inst_timeseries.R
+++ b/2_process/src/munge_inst_timeseries.R
@@ -70,6 +70,7 @@ aggregate_data_to_daily <- function(inst_data, daily_data){
   only_inst_data = setdiff(inst_data$site_no, daily_data$site_no)
   
   daily_values <- p1_inst_data %>%
+    filter(site_no %in% only_inst_data) %>%
     mutate(Date = as.Date(dateTime, format="%Y-%m-%d")) %>%
     group_by(Date, site_no) %>%
     summarise(Value = mean(Value_Inst), 
--- a/2_process/src/munge_inst_timeseries.R
+++ b/2_process/src/munge_inst_timeseries.R
@@ -95,7 +95,7 @@ aggregate_data_to_daily <- function(inst_data, daily_data, min_daily_coverage, o
               Value_Min_cd = first(Value_Inst_cd),
               .groups="keep") %>%
     mutate(percent_coverage=value_count/(value_count + na_count)) %>%
-    filter(percent_coverage > min_daily_coverage) %>%
+    filter(percent_coverage >= min_daily_coverage) %>%
     select(-c(na_count, value_count, percent_coverage))
   
   return(daily_values)
--- a/2_process/src/munge_inst_timeseries.R
+++ b/2_process/src/munge_inst_timeseries.R
@@ -55,7 +55,9 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
 
 }
 
-aggregate_data_to_daily <- function(inst_data, daily_data, output_tz){
+
+
+aggregate_data_to_daily <- function(inst_data, daily_data, min_daily_coverage, output_tz){
   #' 
   #' @description Function to aggregate instantaneous NWIS data collected at sub-hourly (e.g. 15/5/30 min) intervals to hourly min/mean/maxs
   #'
@@ -63,6 +65,11 @@ aggregate_data_to_daily <- function(inst_data, daily_data, output_tz){
   #' inst_data must include the following columns: c("Value_Inst","dateTime","site_no")
   #' @param daily_data a data frame the downloaded daily time series of DO data. This is used so that if a site is already in 
   #'  the daily sites, we won't do the aggregating here
+  #' @param min_daily_coverage - float 0-1, minimum coverage needed to return daily summaries.
+  #' A min_daily_coverage of 0.5 means that summary stats will be calculated for 
+  #' days with at least 50% coverage. Coverage is calculated on a daily basis
+  #'  as (num_non_na_vals/(num_non_na_vals + num_na_vals))
+  #' 
   #' @param output_tz character string to set display attribute of dateTime. Possible values to provide are "UTC",
   #' "America/New_York","America/Chicago", "America/Denver","America/Los_Angeles", "America/Anchorage", 
   #' as well as the following which do not use daylight savings time: "America/Honolulu", "America/Jamaica",
@@ -77,14 +84,15 @@ aggregate_data_to_daily <- function(inst_data, daily_data, output_tz){
     filter(site_no %in% only_inst_data) %>%
     mutate(dateTime_local = lubridate::with_tz(dateTime,tzone=output_tz),
            Date = lubridate::date(dateTime_local)) %>%
-    group_by(Date, site_no, .groups = "keep") %>%
+    group_by(Date, site_no, agency_cd, Parameter) %>%
     summarise(Value = mean(Value_Inst, na.rm=TRUE), 
               Value_Min = min(Value_Inst, na.rm=TRUE), 
               Value_Max = max(Value_Inst,na.rm=TRUE), 
               na_count=sum(is.na(Value_Inst)), 
               value_count=sum(!is.na(Value_Inst))) %>%
     mutate(percent_coverage=value_count/(value_count + na_count)) %>%
-    filter(percent_coverage > 0.5)
+    filter(percent_coverage > min_daily_coverage) %>%
+    select(-c(na_count, value_count, percent_coverage))
   
   return(daily_values)
 }
--- a/2_process/src/munge_inst_timeseries.R
+++ b/2_process/src/munge_inst_timeseries.R
@@ -55,7 +55,7 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
 
 }
 
-aggregate_data_to_daily <- function(inst_data, daily_data){
+aggregate_data_to_daily <- function(inst_data, daily_data, output_tz){
   #' 
   #' @description Function to aggregate instantaneous NWIS data collected at sub-hourly (e.g. 15/5/30 min) intervals to hourly min/mean/maxs
   #'
@@ -63,15 +63,20 @@ aggregate_data_to_daily <- function(inst_data, daily_data){
   #' inst_data must include the following columns: c("Value_Inst","dateTime","site_no")
   #' @param daily_data a data frame the downloaded daily time series of DO data. This is used so that if a site is already in 
   #'  the daily sites, we won't do the aggregating here
+  #' @param output_tz character string to set display attribute of dateTime. Possible values to provide are "UTC",
+  #' "America/New_York","America/Chicago", "America/Denver","America/Los_Angeles", "America/Anchorage", 
+  #' as well as the following which do not use daylight savings time: "America/Honolulu", "America/Jamaica",
+  #' "America/Managua","America/Phoenix", and "America/Metlakatla"
   #'
   #' @value A data frame containing daily min, mean, and max values the parameter of interest
   
   
   only_inst_data = setdiff(inst_data$site_no, daily_data$site_no)
   
-  daily_values <- p1_inst_data %>%
+  daily_values <- inst_data %>%
     filter(site_no %in% only_inst_data) %>%
-    mutate(Date = as.Date(dateTime, format="%Y-%m-%d")) %>%
+    mutate(dateTime = with_tz(dateTime,tzone=output_tz)) %>%
+    mutate(Date = date(dateTime)) %>%
     group_by(Date, site_no) %>%
     summarise(Value = mean(Value_Inst), 
               Value_Min = min(Value_Inst), 
--- a/2_process/src/munge_inst_timeseries.R
+++ b/2_process/src/munge_inst_timeseries.R
@@ -90,6 +90,9 @@ aggregate_data_to_daily <- function(inst_data, daily_data, min_daily_coverage, o
               Value_Max = max(Value_Inst,na.rm=TRUE), 
               na_count=sum(is.na(Value_Inst)), 
               value_count=sum(!is.na(Value_Inst)),
+              Value_cd = first(Value_Inst_cd),
+              Value_Max_cd = first(Value_Inst_cd),
+              Value_Min_cd = first(Value_Inst_cd),
               .groups="keep") %>%
     mutate(percent_coverage=value_count/(value_count + na_count)) %>%
     filter(percent_coverage > min_daily_coverage) %>%
--- a/2_process/src/munge_inst_timeseries.R
+++ b/2_process/src/munge_inst_timeseries.R
@@ -89,7 +89,8 @@ aggregate_data_to_daily <- function(inst_data, daily_data, min_daily_coverage, o
               Value_Min = min(Value_Inst, na.rm=TRUE), 
               Value_Max = max(Value_Inst,na.rm=TRUE), 
               na_count=sum(is.na(Value_Inst)), 
-              value_count=sum(!is.na(Value_Inst))) %>%
+              value_count=sum(!is.na(Value_Inst)),
+              .groups="keep") %>%
     mutate(percent_coverage=value_count/(value_count + na_count)) %>%
     filter(percent_coverage > min_daily_coverage) %>%
     select(-c(na_count, value_count, percent_coverage))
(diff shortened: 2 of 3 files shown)


2_process/src/munge_inst_timeseries.R
```diff
@@ -55,3 +55,26 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
...
+  #'
+  #' @param inst_data a data frame containing the downloaded time series for NWIS instantaneous site. 
+  #' inst_data must include the following columns: c("Value_Inst","dateTime","site_no")
+  #' @param daily_data a data frame the downloaded daily time series of DO data. This is used so that if a site is already in 
+  #'  the daily sites, we won't do the aggregating here
+  #'
+  #' @value A data frame containing daily min, mean, and max values the parameter of interest
+  
+  
+  only_inst_data = setdiff(inst_data$site_no, daily_data$site_no)
```
lekoenig, line 70: Here we're only calculating daily summaries for instantaneous sites that don't report daily summaries (n=15 SC sites right now). That makes sense, since I think we'll want to default to the reported daily values whenever they are available. However, each site would have to report each of the stat_cd's we're interested in - mean/max for inland-salinity. It looks like most of the daily sites do report mean/min/max but at least one SC site (01473169) only reports mean/min. So we would either forgo the max data for that site, or calculate the daily summaries and handle the uv/dv selection in your `p2_daily_combined` target. Do you have any thoughts on that?

2_process/src/munge_inst_timeseries.R
```diff
@@ -55,3 +55,26 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
...
+  
+  daily_values <- inst_data %>%
+    filter(site_no %in% only_inst_data) %>%
+    mutate(Date = as.Date(dateTime, format="%Y-%m-%d")) %>%
```
lekoenig, line 74: `as.Date` will return the date associated with the time zone of `dateTime`, which is currently "UTC". For daily summaries I think we should consider using local time to aggregate the data to daily time steps (some discussion of time zones from the drb-do project is [here](https://github.com/USGS-R/drb-do-ml/issues/22). I would guess the local WSC's also report the daily data using dates based on local time.

If you agree with using local time, one way to implement that is to use helper functions in the {lubridate} package:
<pre>
dateTime_local = lubridate::with_tz(dateTime,tzone="America/New_York")
Date = lubridate::date(dateTime_local)
</pre>

2_process/src/munge_inst_timeseries.R
```diff
@@ -55,3 +55,26 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
...
+    group_by(Date, site_no) %>%
+    summarise(Value = mean(Value_Inst), Value_Min = min(Value_Inst), Value_Max = max(Value_Inst)) %>%
+    na.omit()
```
lekoenig, line 77: Did you add `na.omit()` because `NA` values were being returned. For the functions `mean()`, `min()`, etc., they will return `NA` if there are any NA values in the input vector unless you specify to remove them, e.g. `mean(x,na.rm=TRUE)`. However, NA's are currently filtered out in the fetch/data download step so I wouldn't expect the additional argument to make a difference here.

lekoenig:
Do we want to do any filtering or flagging of the instantaneous data that inform daily summaries? For example, at least x% of the unit values should be present to calculate a mean/min/max? This is often the reason that single days are missing from the reported daily values on NWIS, for example.

2_process.R
```diff
@@ -30,5 +30,15 @@ p2_targets_list <- list(
...
+  
+  # Aggregate instantaneous DO data to daily min/mean/max
+  tar_target(
+    p2_inst_data_daily,
+    aggregate_data_to_daily(p1_inst_data,p1_daily_data)),
+  
+  # Combine 1) daily DO data and 2) instantaneous DO data that has been aggregated to daily 
+  tar_target(
+    p2_daily_combined,
+    bind_rows(p1_daily_data, p2_inst_data_daily))
```
lekoenig, line 43: Thanks for initializing this target that combines the NWIS continuous datasets. Do we still want to pad the timeseries to format for modeling (per discussion in #15)? If so, it might make sense to do that formatting within this target. What do you think?

2_process/src/munge_inst_timeseries.R
jds485: Yes, good call on aggregating in local time. I think that's the way to catch daylight savings and leap year quirks that happen in some datasets (looks like your DO data had a site like that)

jds485:
> Do we want to do any filtering or flagging of the instantaneous data that inform daily summaries? For example, at least x% of the unit values should be present to calculate a mean/min/max? This is often the reason that single days are missing from the reported daily values on NWIS, for example.

Do you know the rule used by NWIS? It would be great to adopt the same rules

lekoenig:
> Do you know the rule used by NWIS? It would be great to adopt the same rules

It seems like the local user/data maintainer can use discretion based on site knowledge and streamflow conditions, but basic guidelines for SC are that at least 50% of the unit values are present to publish the daily min/max (for DO this is 25%). No specific guidelines are given for reporting the mean other than to say that it's a judgement call when less than 100% of the expected unit values are present. (My reference for these sort of questions has been this 2006 [technical report](https://nc.water.usgs.gov/reports/tm1D3/pdf/TM1D3.pdf) on SOPs for continuous water quality monitoring.)

The number of daily expected unit values is not constant across sites or sometimes even within a site (e.g. a site might record for 10 min for 2 years, then switch to 15 min for 1 year, then switch to 5 min intervals for a period...). So to add this flagging step we'd need to edit the data download functions to retain NA values.

2_process/src/munge_inst_timeseries.R
lekoenig: I also compared daily summaries based on UTC vs. local eastern time for a few sites that report instantaneous data as well as the daily summaries. The local time estimates were a better match, so I think it's safe to go with that approach.

jds485:
>The number of daily expected unit values is not constant across sites or sometimes even within a site (e.g. a site might record for 10 min for 2 years, then switch to 15 min for 1 year, then switch to 5 min intervals for a period...). So to add this flagging step we'd need to edit the data download functions to retain NA values.

Could we work with the hourly aggregated timeseries instead of the raw data? Then we can say that we require 12/24 to report values. The max and min would correspond to max/min of hourly average SC, so a little different of an interpretation, but that may help to remove noise within the instantaneous max/min data, anyway.

2_process/src/munge_inst_timeseries.R
jsadler2: I think that the padding can happen downstream of this step. At least in past work it has happened downstream, closer to where the actual dynamic modeling happens.

2_process/src/munge_inst_timeseries.R
jsadler2: Oh okay. Good question. I found this code from another project and it had it there, but I did not actually test whether it was doing anything. I suppose, though, if we are leaving the NA values in so we can calculate the percentage of daily coverage, we'll need to handle those here.

jsadler2:
@jds485, @lekoenig - I might have missed this in a previous conversation/decision - do we not care about daily _minimums_ (just looking at means/maxs)? I ask b/c in the `p1_daily_data` we only have the `Value` and `Value_Max`.

2_process.R
jsadler2: I think that the padding can happen downstream of this step. At least in past work it has happened downstream, closer to where the actual dynamic modeling happens.

2_process/src/munge_inst_timeseries.R
```diff
@@ -55,3 +55,31 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
...
+    filter(site_no %in% only_inst_data) %>%
+    mutate(Date = as.Date(dateTime, format="%Y-%m-%d")) %>%
+    group_by(Date, site_no) %>%
+    summarise(Value = mean(Value_Inst), 
+              Value_Min = min(Value_Inst), 
+              Value_Max = max(Value_Inst), 
+              na_count=sum(is.na(Value_Inst)), 
+              value_count=sum(!is.na(Value_Inst))) %>%
+    mutate(percent_coverage=value_count/(value_count + na_count)) %>%
+    filter(percent_coverage > 0.5)
```
jsadler2, line 82: Does this look like the right way to be filtering these? And if so, is a 50% daily coverage what we want do go with?

jsadler2: >Does this look like the right way to be filtering these? 

btw, I did a spot check on this and it was what I was expecting, but I just want to make sure I didn't overlook anything simple

2_process/src/munge_inst_timeseries.R
lekoenig: We ultimately want daily mean/max/(min) for all continuous sites. We could take two approaches: 1) assume that the local data contributors have already QA/QC'ed the daily records. To combine the daily-aggregated uv values with the daily data in `p1_daily_data` we could then just take a diff approach: use daily summaries for all dv sites and then just add our inst daily summaries for the ~15 inst only sites. 2) Alternatively, we could try to fill in any gaps in the daily record with our daily summaries but that means assuming the reporting offices didn't leave that gap there for a good reason. 

@jds485 do you have an opinion on that? I think Jeff and I are leaning toward the first option, but we thought you may want to weigh in as well.

2_process/src/munge_inst_timeseries.R
lekoenig: The code looks looks good to me!

2_process/src/munge_inst_timeseries.R
jds485: Just so to make sure I'm understanding the suggestion, daily records = daily summaries = daily mean, min, and max? Did you say earlier this week that you compared dv daily records to equivalent obtained from the uv data and found them to be very similar? If so, approach 1 sounds good to me. 

As an extension of that logic, if we trust the QAQC is complete for dv, then should we drop uv data on the days there are gaps in the daily record? For example, if a day has a complete uv record, it would get a dv entry if the uv data were good. So a missing dv entry would indicate that something was wrong with the uv data and we shouldn't use it. What do you think?

jsadler2: With approach 1, we wouldn't even be calculating daily summaries for `dv` sites, just for `uv` sites. So there's no chance, if the code is doing what I think it is, that we would be "filling in" missing `dv` data with `uv` summaries.

jds485: Also, did you see my [comment](https://github.com/USGS-R/drb-inland-salinity-ml/pull/26#issuecomment-995273276) below about using hourly aggregated data instead of uv? It looks like you've retained the NAs, so it may not be a helpful suggestion anymore

2_process/src/munge_inst_timeseries.R
lekoenig: Doh! @jsadler2 I missed something simple in the lines above this code snipped. We need to add `na.rm=TRUE` as an argument to the `min`, `max`, and `mean` functions. Otherwise, if there are any NA values for a uv interval, the daily summary will return NA.

2_process/src/munge_inst_timeseries.R
lekoenig: Yeah, we decided to aggregate the uv data instead of the hourly-aggregated data because 1) the local data contributors use the uv values when reporting the daily summaries, so using the uv data would be more comparable with the dv data, and 2) because we thought using the hourly-aggregated values would potentially misrepresent the daily extremes (min/max) that we're interested in.

2_process/src/munge_inst_timeseries.R
jsadler2: Thanks for catching that! It's been adjusted now.

2_process/src/munge_inst_timeseries.R
jds485: Okay, that makes sense to use uv peaks, then.

for 2, yes I agree. Using hourly data is a different interpretation (maximum of hourly averages instead of daily maximum), and that's less useful for matching to the WQ criteria. This got me thinking about over what timescale SC and DO change. Pulse road salt events will change SC rapidly. Are there similar pulse events for DO? No need to answer here, just something to be thinking about - are our sampling frequencies matching the timescales of expected WQ changes?

lekoenig: > Just so to make sure I'm understanding the suggestion, daily records = daily summaries = daily mean, min, and max? Did you say earlier this week that you compared dv daily records to equivalent obtained from the uv data and found them to be very similar? If so, approach 1 sounds good to me.

Oof, too many descriptors, but yes I'm using those terms equivalently here. I did previously compare the dv daily records to the equivalent values calculated from the uv data (for the 80 sites that have inst and daily data) and they are very similar: 

compare_uv_and_dv

jsadler2:
@lekoenig and @jds485 - are you good with me merging this?

2_process/src/munge_inst_timeseries.R
jds485: Could make the % cutoff a function argument, but 50% as a threshold sounds good. It does make me wonder if we should require a certain number of samples, too. If we had 2 samples, they would be our max and min values.  

For something like DO, it could help to report the time of the max and min.

jds485: Time is not reported in the dv data (`p1_daily_data`) for salinity, but I wouldn't expect time of day to be important for inland salinity. For tides it could be

2_process/src/munge_inst_timeseries.R
```diff
@@ -55,3 +55,36 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
...
+  only_inst_data = setdiff(inst_data$site_no, daily_data$site_no)
+  
+  daily_values <- inst_data %>%
+    filter(site_no %in% only_inst_data) %>%
+    mutate(dateTime = with_tz(dateTime,tzone=output_tz)) %>%
+    mutate(Date = date(dateTime)) %>%
+    group_by(Date, site_no) %>%
+    summarise(Value = mean(Value_Inst, na.rm=TRUE), 
+              Value_Min = min(Value_Inst, na.rm=TRUE), 
+              Value_Max = max(Value_Inst,na.rm=TRUE),
```
lekoenig, line 83: @jsadler2 I just tried running the code and ran into warnings when all values for a day are NA. Do you have that problem, too?

This can happen if a site reports multiple time series/sensor locations within the river and one of those begins before the either of the time series we select. I see this in site 01434498, for example. One way to get around this problem is to calculate `na_count`, `value_count`, and `percent_coverage` and filter by percent_coverage before calculating the mean/max/min.

2_process/src/munge_inst_timeseries.R
```diff
@@ -55,3 +55,36 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
...
+  #' "America/Managua","America/Phoenix", and "America/Metlakatla"
+  #'
+  #' @value A data frame containing daily min, mean, and max values the parameter of interest
+  
+  
+  only_inst_data = setdiff(inst_data$site_no, daily_data$site_no)
+  
+  daily_values <- inst_data %>%
+    filter(site_no %in% only_inst_data) %>%
+    mutate(dateTime = with_tz(dateTime,tzone=output_tz)) %>%
```
lekoenig, line 78: I'd suggest the following minor changes for code readability:

- throughout the repo, I've tried to indicate the function source when we're using packages outside of base and tidyverse. We could indicate the source of the time parsing functions with `lubridate::with_tz(...)` and `lubridate::date()`
- the two calls to `mutate` can be combined into one call, with the two new variables separated by a comma
- I'd suggest replacing `dateTime` with `dateTime_local` or something similar that more readily communicates that we're changing the time zone for the daily aggregation (rather than writing over the original time stamps)

2_process/src/munge_inst_timeseries.R
lekoenig: > Could make the % cutoff a function argument

I agree that the percent coverage cutoff should be a function argument since the USGS guidance varies among some constituents (e.g. for DO, it is 25%). 

> For something like DO, it could help to report the time of the max and min.

I agree that knowing the timing of daily min/max may help inform the processes that are driving daily variation in DO. We would want to know that information for all instantaneous sites rather than just the subset of "inst only" sites, so if we didn't want to include here, we could either create a separate target, or add timing in later when we are working on the dynamic models.

2_process/src/munge_inst_timeseries.R
```diff
@@ -55,3 +55,36 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
...
+    mutate(Date = date(dateTime)) %>%
+    group_by(Date, site_no) %>%
+    summarise(Value = mean(Value_Inst, na.rm=TRUE), 
+              Value_Min = min(Value_Inst, na.rm=TRUE), 
+              Value_Max = max(Value_Inst,na.rm=TRUE), 
+              na_count=sum(is.na(Value_Inst)), 
+              value_count=sum(!is.na(Value_Inst))) %>%
```
lekoenig, line 85: Can add `.groups = "keep"` to the summarize call to get rid of the warning message

jsadler2: Odd. I haven't seen any warnings. I do see this though when I build with `.groups = "keep"`:
```R
* start target p2_inst_data_daily
`summarise()` has grouped output by 'Date', 'site_no'. You can override using the `.groups` argument.
```
w/o that I get 
```R
* start target p2_inst_data_daily
`summarise()` has grouped output by 'Date'. You can override using the `.groups` argument.
```

2_process.R
```diff
@@ -38,5 +38,15 @@ p2_targets_list <- list(
...
+  
+  # Aggregate instantaneous DO data to daily min/mean/max
+  tar_target(
+    p2_inst_data_daily,
+    aggregate_data_to_daily(p1_inst_data,p1_daily_data,"America/New_York")),
+  
+  # Combine 1) daily DO data and 2) instantaneous DO data that has been aggregated to daily 
+  tar_target(
+    p2_daily_combined,
+    bind_rows(p1_daily_data, p2_inst_data_daily))
```
lekoenig, line 51: `bind_rows` will match shared column names, but in this case the data frames `p1_daily_data` and `p2_inst_data_daily` don't share the same structure, so it just fills in NA's when variables aren't available, including Value_Min for all daily sites:

same_input

Adding 'agency_cd' and 'Parameter' to the `group_by` call within `aggregate_data_to_daily()` would save these columns in the summarized `p2_inst_data_daily` data frame. We could also add a `select` step to the filtering/aggregating pipe to omit `na_count` and `value_count` since we don't have those variables for the "dv" data. 

Do we want to retain the data quality codes? If so, we can create those within the mungeing step. If not, we should be able to use `bind_rows(p1_daily_data[,-c(Value_cd,Value_Max_cd)],p2_inst_data_daily)`

2_process/src/munge_inst_timeseries.R
lekoenig: I guess it's not a warning, but just a message from {dplyr}. I'm surprised you see those messages when you build with .groups="keep". Out of curiosity do you have a comma in between the last command and .groups = "keep" (`value_count=sum(!is.na(Value_Inst)), .groups="keep"`)

jsadler2: I have this:
```R
group_by(Date, site_no, .groups = "keep") %>%
```

lekoenig: Oh, sorry I realize my original comment wasn't very clear. `.groups="keep"` is an argument within the `summarize` step.

lekoenig:
```
summarize(Value = mean(Value_Inst,na.rm=TRUE), 
              Value_Min = min(Value_Inst,na.rm=TRUE), 
              Value_Max = max(Value_Inst,na.rm=TRUE), 
              na_count=sum(is.na(Value_Inst)), 
              value_count=sum(!is.na(Value_Inst)),
              .groups="keep")
```

2_process/src/munge_inst_timeseries.R
lekoenig: I realized I got this error when running and testing the daily aggregation steps on the full `p1_inst_data`. So perhaps this isn't an issue for the data subset returned by the `setdiff` step.

2_process/src/munge_inst_timeseries.R
jsadler2: No problem. Thanks for the correction. That just shows my newbie-ness in R :)

So I added that in there and now I don't get any message at all.

Do you mind briefly explaining why that is? What does `.groups = "keep"` actually even do?

2_process.R
jsadler2: I've aligned all of the columns except `the Value_cd` and `Value_Max_cd`. Is it a problem to have `NA` for the `Value_cd` and `Value_Max_cd`? I think I'd prefer to leave those than take out the information that is in those columns for the `dv` sites.

If we feel like we need a code, we'd have to decide what to do if there are multiple values per day. If all of the codes for one day are `A` except one which is `P` - what is the single code for that day?

2_process/src/munge_inst_timeseries.R
lekoenig: `.groups` controls the output structure of the data frame, so specifying `.groups="keep"` results in a "grouped tibble" of class `"grouped_df"` (i.e. the grouping is retained). I pretty much use it to silence the output message from dplyr. We could also specify `.group="drop"` to return an ungrouped tibble/data frame.

2_process.R
lekoenig: > I've aligned all of the columns except the Value_cd and Value_Max_cd in e1f8ac2.

Great, thanks!

> Is it a problem to have NA for the Value_cd and Value_Max_cd? I think I'd prefer to leave those than take out the information that is in those columns for the dv sites.

I think it depends on whether we would potentially want this information in a downstream/modeling step. Do you think there's a use case for that?

In terms of assigning a code, I wonder if it comes up often that there are multiple code values in a single day - from a QA/QC perspective I would guess that the data would get approved in whole-day increments, but I see your point! If we want a code, I wouldn't worry too much about multiple values per day and would probably just take the first one.

2_process/src/munge_inst_timeseries.R
jsadler2: Ah. That makes sense. Thanks for explaining.

2_process/src/munge_inst_timeseries.R
jsadler2: I really like the idea of including this information. I've played around a little bit to calculate these, but haven't found a super clean solution. Given that, and since we have no immediate plans to use this information, I suggest that I open a new issue for this so we don't forget about it. That way we can move forward with merging what we have for calculating the min/mean/maxs and leave the calculating of the min/max timings to another PR. Does that sound good, @lekoenig, @jds485 ?

lekoenig: OK, glad we're all in agreement that the timing data could potentially be useful (and thanks for taking a first stab at it, @jsadler2!). I'm fine with moving forward and leaving the timing for another PR. I have some ideas to try and could look into this further if we have an issue for it.

jds485: Sounds good to me to move timing data to a new issue

2_process.R
jds485: Could be helpful to for diagnostics of model performance (e.g., do days with mostly P values have worse performance than days with mostly A?). A daily-aggregated "code" could be the proportion of samples that are not A.

2_process/src/munge_inst_timeseries.R
jsadler2: 👍 - here it is: https://github.com/USGS-R/drb-inland-salinity-ml/issues/34

2_process.R
jsadler2: I agree that this could be useful. 

I'm looking at the data and there are 1231 site-days out of 100894 (0.1%) that have  multiple value codes. So if we take the "first" one for those days, we are losing some information. That includes 1107 site-days with `NaN` as the Value_cd, so there's a chance that for those days, we'd end up with a code of `NaN` (if we are taking the first and the first code is `NaN`). This would not be helpful.

jsadler2: Given that it's only 0.1% of the inst. daily summaries that this applies to, I'm in favor of just being aware of this and then moving forward. I ran the code with taking just the first code and these were the numbers:
```R
> summary(as.factor(p2_inst_data_daily$Value_cd))
    A A [0]     P 
 3333     1  2288 
```
So there aren't any `NaN` that show up here.

jsadler2: Another thing that comes up here is the `Value_Max_cd`. I think the "A" job is to get the `Value_Inst_cd` at the max daily value for this. Since that is basically the same functionality as what we'll need for calculating the "time at max" and that will be addressed in #34, I'd be in favor of leaving it to that issue/PR.

jsadler2: Otherwise for a stand-in, we could just take the first again, knowing that in 99.9% of the site-days that is the same throughout the day.

jds485: Thanks for checking that, Jeff. Sounds like a good approach to handle the `Value_Max_cd` with the issue you cited

lekoenig: Thanks for looking into that more closely. It makes sense that in aggregating to daily time steps we would be losing some information, retaining the first value was just my initial thought on a way to summarize the quality of the uv data . Perhaps taking the last value is even more conservative, since timestamps later in the day are more likely to be tagged as "P" compared to the first value, but it probably isn't too important, since the difference spans such a small proportion of the inst. daily summaries. 

I'm curious, though - I don't see any `NaN` values for `Value_Inst_cd` in `p1_inst_data.` Were you referring to `NA` values, or are there rows where the quality code == `NaN`? If so, it seems like something wonky might be happening with the summarization.

jsadler2: When the `Value_Inst` is `NA` the `Value_Inst_cd` also shows ups as `NA`:
image

lekoenig: OK, yep that makes sense! I thought that `NaN` values were being returned rather than just `NA`

jsadler2:
Alright, @jds485 and @lekoenig - I _think_ I resolved all of the comments/questions. Let me know if you see anything else that I should do before merge.

2_process.R
```diff
@@ -38,5 +38,15 @@ p2_targets_list <- list(
      {
        sites_tbl <- read_csv(p2_site_list_csv)
        get_site_flowlines(p1_reaches_sf, sites_tbl, sites_crs = 4269, max_matches = 1, search_radius = 0.1)
-     })
+     }),
+  
+  # Aggregate instantaneous DO data to daily min/mean/max
```
lekoenig, line 43: Do you mind editing the descriptions here to replace "DO" references with "SC" (and on line 48)?

jsadler2: Haha good catch

jsadler2:
```suggestion
  # Aggregate instantaneous SC data to daily min/mean/max
```

2_process/src/munge_inst_timeseries.R
```diff
@@ -55,3 +55,48 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
...
+              Value_Min = min(Value_Inst, na.rm=TRUE), 
+              Value_Max = max(Value_Inst,na.rm=TRUE), 
+              na_count=sum(is.na(Value_Inst)), 
+              value_count=sum(!is.na(Value_Inst)),
+              Value_cd = first(Value_Inst_cd),
+              Value_Max_cd = first(Value_Inst_cd),
+              Value_Min_cd = first(Value_Inst_cd),
+              .groups="keep") %>%
+    mutate(percent_coverage=value_count/(value_count + na_count)) %>%
+    filter(percent_coverage > min_daily_coverage) %>%
```
lekoenig, line 98: Could change to `filter(percent_coverage >= min_daily_coverage)` if we wanted to include days with _at least_ 50% coverage, but I doubt this minor change would impact many rows (if any at all)

2_process/src/munge_inst_timeseries.R
```diff
@@ -55,3 +55,48 @@ aggregate_data_to_hourly <- function(inst_data,output_tz){
...
+  #' @value A data frame containing daily min, mean, and max values the parameter of interest
+  
+  
+  only_inst_data = setdiff(inst_data$site_no, daily_data$site_no)
+  
+  daily_values <- inst_data %>%
+    filter(site_no %in% only_inst_data) %>%
+    mutate(dateTime_local = lubridate::with_tz(dateTime,tzone=output_tz),
+           Date = lubridate::date(dateTime_local)) %>%
+    group_by(Date, site_no, agency_cd, Parameter) %>%
```
lekoenig, line 87: I noticed that the table that gets returned (e.g. by `tail(p2_inst_data_daily)`) looks a bit odd since it is sorted first by Date, and then by site number. The values should not change, but if we wanted to aesthetically match the p1_daily_data I think you should just be able to swap the order of `Date` and `site_no` here.

2_process/src/munge_inst_timeseries.R
jsadler2: to me `min_daily_coverage` means that if I have at least that daily coverage it will be included. So I think it should be `>=` like you suggest, instead of `>`.
```suggestion
    filter(percent_coverage >= min_daily_coverage) %>%
```

2_process/src/munge_inst_timeseries.R
jsadler2: good idea
```suggestion
    group_by(site_no, Date, agency_cd, Parameter) %>%
```

lekoenig:
Thanks for taking on the daily summaries, Jeff! The pipeline builds for me. I made a couple small formatting comments/suggestions, so you can incorporate them if they resonate with you, or leave as is. I'm not sure if @jds485has is planning on running the code, but I approve the PR and am happy to merge.

jds485:
I'm not going to run the code for this PR. I reviewed the diff of changes and they look good to me

PR #31, opened 2021-12-14
Create log/ind files to track data changes

### LINKED ISSUE #23: Add log and summary files to track changes to SC data
Add targets that will help us track changes in the data through time: 
1. output a log/indicator file containing a data hash for each site - the log file is mostly used for pipelines-purposes, and to decide when to push updated data to our shared data cache (e.g. S3 bucket)
2. output a human-readable summary file that contains information about data changes (e.g. number of sites, number of observations, perhaps summary stats like median, min, max). Some of this information is currently contained in the `3_visualize/out/report-wqp-salinity-data.html` but should be moved to a format that can be committed and tracked through diffs

### LINKED ISSUE #29: NWIS sites added (and maybe removed?) over time
While rebuilding targets for PR review #27, I learned that NWIS sites can be added (and maybe also removed?) from one day to the next, which raises reproducibility concerns. We will need a way to handle tracking exactly which sites are pulled to ensure consistency among repo developers and future users.

To elaborate, the new site had data for most of the specified date range. So it should have been gathered if it were previously available on NWIS.

lekoenig:
This PR adds special handling for data download for site 01482537 and also creates log/indicator files to track SC data changes over time. There are currently 7 files that get saved in "3_visualize/log" and committed so that we can track diffs using git:

- 2 ind files containing the data hash for daily and instantaneous NWIS sites, respectively 
- 2 summary files containing basic summary statistics (count of observations, mean, sd) for daily and instantaneous NWIS sites, respectively
- 1 site list file that describes the number of unique sites represented
- 1 summary file that indicates NWIS instantaneous "uv" sites that have multiple time series. These sites have special handling for data download.
- 1 ind file containing the data hash for the filtered WQP data (discrete SC data)

I view the ind files as mostly useful for pipeline purposes (including integration with S3) and the summary files as a human-readable summary file, so I've opted to keep them separate for now. I'd be open to any input on that, and hopefully these changes will help us track data changes in the pipeline (including whether changes are due to a data update or to changes in the code). 

closes #23, closes #29


1_fetch.R
```diff
@@ -42,6 +42,17 @@ p1_targets_list <- list(
...
   
+  # Create log file to track sites with multiple time series
+  tar_target(
+    p1_nwis_sites_inst_multipleTS_csv,
+    p1_nwis_sites %>%
+      # retain "uv" sites that contain data records after user-specified {earliest_date}
+      filter(data_type_cd=="uv",!(site_no %in% omit_nwis_sites),end_date > earliest_date) %>%
+      # save record of sites with multiple time series
+      group_by(site_no) %>% mutate(count_ts = length(unique(ts_id))) %>%
+      filter(count_ts > 1) %>%
```
jds485, line 53: should this be `>= 1`? Or is it `> 1` because that indicates multiple timeseries for that site?

3_visualize/log/daily_timeseries_summary.csv
```diff
@@ -0,0 +1,112 @@
+site_no,n_obs,mean,sd
+01411400,2,141,0
+01411500,3158,77.06188511583625,8.128900673491826
```
jds485, line 3: Can probably round to 6 decimal places for mean and sd. There may be some numerical precision differences between users, but out to 6 places should be the same

3_visualize/log/daily_timeseries_summary.csv
```diff
@@ -0,0 +1,112 @@
+site_no,n_obs,mean,sd
+01411400,2,141,0
```
jds485, line 2: Did you only save sites with 2 or more data points? sd would not exist for 1  point, so I'm wondering if all of those sites got omitted from the log file as a result

1_fetch.R
lekoenig: For the `p1_nwis_sites_inst_multipleTS_csv` target I wanted to track sites that return more than one time series id, for example, if sensors are moved or if multiple sensors exist at a site. These sites are the ones that require special handling for the data download steps. Does that make sense?

3_visualize/log/daily_timeseries_summary.csv
lekoenig: Great point, I'll make the rounding change in the `summarize_timeseries.R` file.

3_visualize/log/daily_timeseries_summary.csv
lekoenig: This summary file (`3_visualize/log/daily_timeseries_summary.csv`) is for the daily NWIS data. All daily sites that get returned have at minimum 2 days reported, so I have not done any additional filtering here. 

The discrete data would include examples where there is only 1 data point for a given site. However, I haven't saved/committed a similar summary file for the discrete data since it is static (apart from periodic updates) and less subject to changes.

1_fetch.R
jds485: Yes, that makes sense. Thanks! Were there date overlaps with different values reported for the site? How did you handle those?

lekoenig: Yes, there are often date overlaps for sites with multiple time series. For sites with multiple time series I check the site remarks and inspect the different time series before deciding on a data handling strategy. With this commit I've tried to add some in-line documentation explaining what I found and the site-specific data handling steps (see `get_inst_nwis_data.R`) for each of these sites. 

For some of the sites, the multiple time series represent additional sensors within piezometers, side channels, etc. that are not of primary interest to our project right now. So the data-handling lines simple select the time series of interest that is representative of conditions in the main river channel.

For at least one site ([01467200](https://waterdata.usgs.gov/nwis/uv?site_no=01467200)), overlapping sensor deployments indicate good comparability between the two time series during the period of simultaneous measurement. So in that case, we merge the two time series by accepting values from the "new" time series when available and accepting values from the "old" time series if not.

3_visualize/log/daily_timeseries_summary.csv
jds485: Does discrete data here mean qw from NWIS, or are there other sources? Is there a non-timeseries summary for these discrete data or are they otherwise tracked? For example, we could add total number of discrete samples and sites to `3_visualize/log/sitelist_summary.csv`.

jds485 (changes requested):
Looks good and ran for me! I've requested rounding in log files. Ongoing discussion about committing info for discrete samples.

3_visualize/log/daily_timeseries_summary.csv
lekoenig: Yes, discrete data would mean "qw" data for USGS-sites, but for the sake of our project discrete data means data from the harmonized WQP dataset (which includes the NWIS water quality data). Right now, the discrete data are tracked with an indicator file w/ a data hash (`3_visualize/log/wqp_data_ind.csv`). I didn't add more summary information from the harmonized WQP data because I don't expect that dataset to be updated very often. 

The number of sites in the filtered WQP dataset is included in `3_visualize/log/sitelist_summary.csv`. Right now we only return the total observation-days across discrete and continuous measurements, but we could add a column that indicates the number of discrete-only observation-days. I'm curious if you were able to make sense of the `sitelist_summary`. Some sites intersect the discrete data, daily data, and instantaneous data, so my approach was to give 'total' numbers as well as numbers that reflect unique site id's/latlon. The `n_allsites_nwis` columns would be the columns to check to see whether any new "dv" or "uv" sites were added in a given pipeline build.

1_fetch.R
jds485: Thanks for doing those checks. All make sense to me. Could average values for overlaps instead of taking new, but I realize that can be a judgement call if the new data are also a more reliable sensor or site location than the old

3_visualize/log/daily_timeseries_summary.csv
jds485: Okay, that makes sense for not tracking discrete harmonized WQP data. I think the data hash is sufficient. 

Yes, I think I followed the logic you used for columns to report in the `sitelist_summary`. It would be helpful to track discrete only observations. This could be important once the updated to 2021 WQP data are pulled into this workflow.

1_fetch.R
lekoenig: I agree that the averaging approach could also make sense for multiple [river channel] time series. The reason I didn't take that approach with site 01467200 is because the sensors generally have different timesteps - the old sensor recorded every 15 min whereas the new sensor records every 5 min. So I figured if the "between-15 min" timestamps would only represent the new sensor anyway, I may as well preferentially select that time series during the period of overlap.

jds485: Wow, 5 minutes! Okay, that sounds good

3_visualize/log/daily_timeseries_summary.csv
lekoenig: Rounding changes made in new commit `2ff9b1f`

jds485 (approved):
Looks good! You can merge when ready

Issue #21, opened 2021-12-06
Create cross-walk table from PRMS segment ID to NHDPlusV2 COMID for DRB


lekoenig:
Develop cross walk table between PRMS and NHD; check matched NHD reach assignments

lekoenig:
We ultimately want to be able to crosswalk between PRMS segments and NHDPlusV2 flowlines to take advantage of rich attribute datasets that are referenced to NHDPlus. Below I've added some notes that describe what that process looks like right now. 

The geospatial fabric (GF)/PRMS network is less dense relative to the NHD, and PRMS reaches are often aggregated versions of NHD flowline segments. The [GF](https://www.sciencebase.gov/catalog/item/535eda80e4b08e65d60fc834) contains a variable called `POI_ID` that corresponds to the NHDPlus COMID at the downstream end of the PRMS reach. Given this schema, we can use a tabular join between GF and the [DRB-PRMS segments](https://www.sciencebase.gov/catalog/item/5f6a285d82ce38aaa244912e) to identify the NHDPlusV2-COMID at the downstream end of each reach, based on the shared segment ID. 

![tabular_join_example](https://user-images.githubusercontent.com/8785034/145833326-71ad6259-5764-48ef-a139-b0ec17e8fecc.png)

However, in inspecting the results I've noticed approximately 20-25 edge cases where the tabular join may not result in the expected PRMS-NHD match. These edge cases generally fall into 3 categories:  

1. As best as I can tell, the GF flowline segments were referenced to NHDPlusV1, necessitating an additional step to crosswalk between the COMID and `POI_ID` from NHDPlusV1 to the NHDPlusV2-COMID. In most cases, these COMID's are a 1-1 match. There are a couple of cases where the crosswalk between NHDPlusV1 and NHDPlusV2 is not 1-1, and so the COMID referenced in `POI_ID` doesn't correspond to the NHDPlusV2-COMID at the bottom of the reach: 

![xwalk_not_1_1](https://user-images.githubusercontent.com/8785034/145845096-dae52c62-c1a6-40e7-ab71-d10ee9ae2117.png)

2. For the DRB-PRMS network, a subset of segments were split such that  segment start and endpoints were connected by one reach with no intersecting reaches. So in a couple of cases, two DRB-PRMS reaches (indicated by unique `subsegid`'s) share the same NHD reach:  

[...]

PR #28, opened 2021-12-10
Fetch NLCD land cover data from NHDPlus v.2.1

msleckman:
Opening a draft PR with the fetch functions for Land Cover Data 

In `get_NLCD_LC.R`, drafted two functions: `download_NHD_NLCD_data()` and `unzip_NHD_NLCD_data()`

Key Elements:
* Downloading using `sbtools` from Land Cover land cover attributes in NHDPlus 
* require sb_id or list of sb_ids
* mapped in subfolder build out and organization
* Unzipping 

To do: 
- [x] Review function above and build target in `1_fetch.R`
- [x] Add read in function --> Update (12-13-21), prefer subseting beforehand as the files at conus scale are too large.

PR #35, opened 2021-12-22
add prms-nhdv2 crosswalk target

lekoenig:
This PR adds two targets to the pipeline that facilitate moving between PRMS and NHDPlusV2 representations of the river network for the purpose of gathering river attributes:

- `p1_nhdv2reaches_sf` retrieves the NHDPlusV2 flowlines for the DRB based on our HUC08 watersheds of interest
- `p2_prms_nhdv2_xwalk` pairs each PRMS segment with its associated NHDPlusV2 flowline reaches, and outputs a data frame that contains the most downstream COMID as well as a vector containing all of the NHD reaches that make up the PRMS segment.

xwalk_table

comids_all_example

I used a spatial join between `p1_nhdv2reaches_sf` and a buffered polygon around the PRMS lines to identify intersecting NHD reaches (discussed in issue #21). In theory, the spatial join would give us all of the contributing NHD reaches (`comid_all` above), however, this didn't work well for cases where the PRMS lines don't map exactly onto the NHD. So to make sure we're gathering all of the NHD reaches, I identified the most upstream/downstream reaches from the spatial join, and then used a recursive function to traverse the NHD network between the upstream/downstream ends. I checked this approach by comparing the length of the PRMS segment with the summed length of the contributing NHD reaches from `comid_all.` There were 6 PRMS segments where these lengths differed by > 1 m, and these seem to occur because PRMS doesn't always map onto the NHD lines, e.g.

nonoverlapping

I'm adding @jsadler2 as a primary code reviewer and @msleckman as a conceptual reviewer (@msleckman, does this format work for gathering the MasterList land cover attributes, for example? Is there anything else we need?). 

FYI, the `p2_prms_nhdv2_xwalk`  takes about 5 minutes to build on my local machine.


1_fetch/src/get_nhdplusv2.R
```diff
@@ -0,0 +1,18 @@
...
+  #'
+  #' @param huc8 string or character vector containing the HUC8 watersheds of interest
+  #'
+  #' @value An sf object containing NHDPlusV2 flowlines
+  
+  huc8_basin <- suppressMessages(nhdplusTools::get_huc8(id=huc8))
+  huc8_bbox <- sf::st_as_sfc(st_bbox(huc8_basin))
+  
+  # Download flowlines within the bounding box of the HUC8 basin
+  # Use the bounding box rather than the basin boundary to make sure we're capturing all of the reaches
```
jsadler2, line 13: Interesting. Did you see cases where some reaches were being missed with the huc polygons (as opposed to the BB)?

1_fetch.R
```diff
@@ -86,6 +87,25 @@ p1_targets_list <- list(
...
     st_read(p1_reaches_shp)
+  ),
+  
+  # Download NHDPlusV2 flowlines for DRB
+  tar_target(
+    p1_nhdv2reaches_sf,
+    {
+      drb_huc8s %>%
+        # For each HUC8 watershed, retrieve NHDPlusV2 flowlines
+        purrr::map(.,get_nhdv2_flowlines) %>%
```
jsadler2, line 98: Curious why you are using `purrr` instead of `targets` dynamic branching.

2_process.R
```diff
@@ -40,13 +51,15 @@ p2_targets_list <- list(
...
-  # Combine 1) daily DO data and 2) instantaneous DO data that has been aggregated to daily 
+  # Pair PRMS segments with NHDPlusV2 reaches
   tar_target(
-    p2_daily_combined,
-    bind_rows(p1_daily_data, p2_inst_data_daily))
+    p2_prms_nhdv2_xwalk,
+    {
+      p1_reaches_sf %>%
+        split(.,.$subsegid) %>%
+        purrr::map(.,pair_nhd_reaches,nhd_lines=p1_nhdv2reaches_sf) %>%
```
jsadler2, line 60: same question as above re `purrr` instead of `targets` dynamic branching.

jsadler2 (approved):
Looks good, @lekoenig. I have a couple very minor things.

For the record:
I was thinking that the longer running time was because of the recursive function (`traverse_nhd`) so I was looking for alternatives (like https://usgs-r.github.io/nhdplusTools/reference/get_DD.html). Then I isolated the two functions and I found that the bulk of the running time actually comes from before that function call - I think it comes from running `sf::st_intersection` which I'm not sure we can get around.

2_process/src/pair_nhd_reaches.R
```diff
@@ -0,0 +1,126 @@
...
+  prms_line_proj <- sf::st_transform(prms_line,5070)
+  nhd_lines_proj <- sf::st_transform(nhd_lines,5070)
+  
+  # Create small buffer (0.3 m) around PRMS segment:
+  prms_line_buffer <- sf::st_buffer(prms_line_proj,dist = 0.3)
+  
+  # Find intersection between NHD reaches and buffered PRMS polygon, and calculate lengths of intersecting NHD reaches
+  lines_int <- suppressWarnings(sf::st_intersection(nhd_lines_proj,prms_line_buffer)) %>%
+    mutate(len = as.numeric(sf::st_length(.)))
+
```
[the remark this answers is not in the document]
lekoenig: I like that suggestion and it cleans things up a bit on line 44.

1_fetch.R
msleckman: If you go through the branching approach, might also be easier to put this process into a separate function and then apply dynamic branching. I am not sure if it would work better, but just a suggestion.

1_fetch/src/get_nhdplusv2.R
lekoenig: I opted to use bounding boxes because I noticed the most downstream PRMS reaches don't get returned if I retrieve NHD by the HUC8basins:

compare

I'm not sure if this really matters for our purposes since we're focused on the upstream segments, but I more or less thought the bbox approach was more conservative in terms of the NHD reaches that get returned.

1_fetch.R
```diff
@@ -86,6 +87,25 @@ p1_targets_list <- list(
   tar_target(
     p1_reaches_sf,
     st_read(p1_reaches_shp)
+  ),
+  
+  # Download NHDPlusV2 flowlines for DRB
+  tar_target(
+    p1_nhdv2reaches_sf,
```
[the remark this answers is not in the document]
msleckman: This is not critical, but one comment about regarding the set-up of targets. From my (new) reviewer perspective, it might be easier to put this piped process into another stand alone function - maybe within `get_nhdv2_flowlines.R`. There are quite a few steps in the dplyr pipe that could be compartmentalized, and also, the function `getnhdv2_flowlines` is a bit hard to catch and review here, since it's within in the pipe (same can be said about the `p2_prms_nhdv2_xwalk` target,  where two functions are called from `pair_nhd_reaches.R`. Just an idea here.

msleckman (commented):
Looks good and full pipeline runs. 😸 
See my comments. Mainly, I have one thought on the calling of functions for building targets, but this pipeline build nicely so not critical. 

Note - this should merge before the [Fetch NLCD PR](https://github.com/USGS-R/drb-inland-salinity-ml/pull/28) because `p1_nhdv2reaches_sf` is an input for the `p1_NLCD_df` target.

msleckman (approved):

1_fetch/src/get_nhdplusv2.R
lekoenig: It looks like I didn't update the legend labels in the above figure, but the panel on the right should read "NHD by bbox"

1_fetch.R
lekoenig: Great points - it does seem like there might be a way to simplify these lines. I had to remind myself that I ended up using `purrr` here for two reasons:  
1) you can only branch over named targets and not global objects. So to branch over `drb_huc8s`, that object would need to be a target. 
2) `nhdplusTools::get_nhdplus` requires an AOI, but seems to have a cap for the number of flowlines that get returned (maybe this is 10,000?). So the population of NHD reaches was getting truncated when I tried to `get_nhdplus()` over the entire DRB. To get around that, I split the DRB into the contributing HUC8 watersheds, and applied the `get_nhdplus` function over each HUC8.

1_fetch.R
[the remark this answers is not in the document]
lekoenig: > From my (new) reviewer perspective, it might be easier to put this piped process into another stand alone function - maybe within get_nhdv2_flowlines.R

Thanks, @msleckman. I agree with your suggestion to move the code currently in the phase file into a separate function. Now all of the manipulation steps occur within `2_process/get_nhdplusv2.R` and the only line within the target is the call to `get_nhdv2_flowlines` Changes are in this commit: https://github.com/USGS-R/drb-inland-salinity-ml/pull/35/commits/629f235be5f8dc980893ba06a68f664f5b806013

2_process.R
lekoenig: I used `purrr` here because as far as I can tell, I cannot apply `targets::map` and dynamic branching to an sf spatial object. I've therefore opted to leave this code for `p2_prms_nhdv2_xwalk` as is, but I can revisit this target if I find out later that branching over spatial objects is possible.

lekoenig:
> For the record:
I was thinking that the longer running time was because of the recursive function (traverse_nhd) so I was looking for alternatives (like https://usgs-r.github.io/nhdplusTools/reference/get_DD.html). Then I isolated the two functions and I found that the bulk of the running time actually comes from before that function call - I think it comes from running sf::st_intersection which I'm not sure we can get around.

Thanks, @jsadler2 for doing these run time tests. I'm not surprised that most of the run time was from `st_intersection` since the `traverse_nhd` function is just querying the attribute table rather than the spatial objects. I briefly investigated the possibility of using `st_intersects` instead - `st_intersects` returns a logical result rather than the intersecting geometries and is much faster than applying `st_intersection` for some use cases. I didn't see a noticeable change in run time for our case though, so I've opted to keep `st_intersection` here.

PR #38, opened 2021-12-28
Match NLCD to PRMS catchments + estimate LC proportions per PRMS catchment

msleckman:
Opening initial PR. Comments and review requests and questions to come. 

Must be merged after Fetch NLCD


jds485:
Did you delete the log files instead of taking the version from USGS main?

msleckman:
@jds485 I did but once this branch is merged to main, the log files will be the ones in main. That seems easiest to me. I suppose for the branch, it's preferred to keep the log content, but I had already deleted the rows in the log files that related to the branch at that point to resolve the merge conflict (per my understanding of our teams discussion then), so I had already made the relevant deletion here.

jds485:
> @jds485 I did but once this branch is merged to main, the log files will be the ones in main.

Do the log files have the same name? Is so, I'd expect that the log files will be deleted in USGS main when this PR is merged because the deletion of log files is within a newer commit than is on USGS main.

lekoenig:
> > @jds485 I did but once this branch is merged to main, the log files will be the ones in main.
> 
> Do the log files have the same name in your repo and in USGS main? If so, I'd expect that the log files will be deleted in USGS main when this PR is merged because the deletion of log files is within a newer commit than is on USGS main.

I agree with Jared's interpretation here. You shouldn't have to delete log files because they either stay the same as in `main` or they get updated in which case we want to commit them.

msleckman:
I see what you mean. And while the log files are indeed in main, they aren't in my branch even if I fetch upstream and sync my branch to main. I will have to incorporate a fix such that the committed deletion doesn't merge to main 😬. Do you have any thoughts on how to proceed?

jds485:
> Do you have any thoughts on how to proceed?
If you re-run your code, the log files should generate. Then you can commit them again.

msleckman:
Ah okay, and so ultimately I am not interested in what is saved in the log file, just that they exist. That's clear. I will rerun to recreate the log files and push this to my branch. 

Now, when merging the PR, how can I ensure that the log files wont merge into main? Do I simply resolve the merge conflict when it arises, and favor of main? In PR it's typically quite difficult to selectively ignore file changes.

jds485:
> how can I ensure that the log files wont merge into main

I'm not expecting a merge conflict because your commits first delete the log files and the add them again. But happy to help if there is a conflict that appears

msleckman:
Alright, this might necessitate a call today possibly. Will keep you posted on teams. 😄

msleckman:
@jds485 - FYI added back the log folder. and True, looks like there is no merge conflict. Likewise for the [older NLCD branch](https://github.com/USGS-R/drb-inland-salinity-ml/pull/28)

2_process.R
```diff
@@ -61,5 +63,40 @@ p2_targets_list <- list(
         purrr::map(.,summarize_paired_comids) %>%
         bind_rows()
     }
-  )
+  ),
+  
+  # Extract COMID and respective area for LC estimation below
```
msleckman, line 68: Added the following 4 targets here for estimating proportions of LC at the PRMS scale.

2_process/src/combine_NLCD_PRMS.R
```diff
@@ -0,0 +1,58 @@
+###-----
+## Subset Land to AOI and join with area and PRMS ids
+AOI_LC_w_area <- function(NLCD_LC_df, area_att, aoi_comids_df){
```
msleckman, line 3: first stand-alone function takes the LC dataset and subsets + joins with PRMS ids from the drb_comids_df

2_process.R
```diff
@@ -61,5 +62,40 @@ p2_targets_list <- list(
...
+  tar_target(p2_nhd_area_att, 
+             p1_nhdv2reaches_sf %>%
+               st_drop_geometry() %>%
+               select(COMID,AREASQKM,TOTDASQKM)
+  ),
+  
+  ## Melt PRMS_nhdv2_xwalk to get all cols of comids Ids filtered to drb 
+  tar_target(p2_drb_comids,
```
msleckman, line 75: Consider simplifying this target and possible combining with the `read_subset()` function when creating `p1_NLCD_df` target

1_fetch/src/get_nlcd_LC.R
```diff
@@ -0,0 +1,161 @@
...
+  }
+  print(list_out_paths)
+  return(list_out_paths)
+}
+
+###-----------------------------------------------------------------------------
+
+read_subset_LC_data <- function(LC_data_folder_path,
+                                Comids_in_AOI_df,
+                                Comid_col = 'comid_down'){
```
msleckman, line 107: this default to be removed. A fix to make in branch:gather_LC_data

msleckman:
Working on heat map showing proportions per PRMS_segid: 

image

jds485:
Land cover [classes](https://www.mrlc.gov/data/legends/national-land-cover-database-2019-nlcd2019-legend). 

In previous projects, I've merged several classes into a single class to improve spatial coverage (e.g., forest = 41, 42, and 43, grass = 51, 71), and have dropped others that do not exist in the region of study. Do we have any perennial ice/snow (class 12)?

msleckman:
@jds485  agreed that would be helpful for our case. I'll incorporate that.

msleckman:
PR ready for review: 

* The added functions in `2_process/src/combine_NLCD_PRMS.R` do the following: 
  * Joins  LC data with area of NHD catchment attributes and with PRMS segids
  * Calculated proportion of each LC class at the level of PRMS 

* A few observations after interpreting and verifying the outputs: 
  * PRMS 287_1 has no area and therefore value was calculated NA (omitted)
  * LC class 12 ([Perennial Ice/Snow](https://www.mrlc.gov/data/legends/national-land-cover-database-2019-nlcd2019-legend)) is not found in the drb for 2011. Omitted for now. 
  
image

See in line comments and questions to guidance in this review.

2_process/src/combine_NLCD_PRMS.R
```diff
@@ -0,0 +1,63 @@
...
+      across(starts_with(paste0('PRMS_AREA_',catchment_att)), ~(.x/AREASQKM_PRMS),
+             .names="PERCENT_{col}"),
+      # round the new cols 
+      across(starts_with(c('PERCENT','PRMS')), round, 2), 
+    .groups="drop")
+  
+  ## Final cleaning - removing LC categories and PRMS_segis that are empty and the empty PRMS (287_1) 
+  area_df <- area_df %>%
+    filter(AREASQKM_PRMS > 0) %>%
+    select(~contains('NLCD11_12'))
```
msleckman: here I am specifically calling out the the LC class 12, but this should be more reproducible in the event that another class might not be present in the given aoi. Do you have suggestions on how to catch that? I was planning to write out something like `select(where(~sum(.) != 0))` however running into col type error as I need to select only the numeric columns for this to work.

jds485:
> * PRMS 287_1 has no area and therefore value was calculated NA (omitted)

This makes me wonder if the NHD attributes for this reach were computed correctly. Are other attributes for this reach also NA or 0? I see that this reach is one of the shortest, so maybe the DEM used to estimate area was not fine enough to resolve its area.

msleckman:
> This makes me wonder if the NHD attributes for this reach were computed correctly. Are other attributes for this reach also NA or 0? I see that this reach is one of the shortest, so maybe the DEM used to estimate area was not fine enough to resolve its area.

@jds485 I agree! And this is the only PRMS in the drb that has a value of 0 for `AREASQKM`. In further investigating, turns out `PRMS_segid:287_1` contains only 1 comid (`COMID:4188275`), and this comid has a `AREASQKM = 0`
```
comid_4188275 <- get_nhdplus(comid = 4188275)
comid_4188275$areasqkm
[1] 0
```
Additionally, could this COMID be erroneous?  The segment does not perfectly align with it's existing stream when mapped. 
```
mapView(comid_4188275 %>% st_transform(3857), crs = 3857)
```
image

This may simply be due to the poor basemap alignment with small streams. Many COMID segments that have valid area are a little off the basemap stream when displayed. 

Finally, There are actually about 47 COMIDs in our drb subset that have 0 value for `areasqkm` (copied full df below if you want to explore). All other COMIDS with 0 area belong to a `PRMS_segid` that contains more than 1 COMID, so those `PRMS_segid` have a area value > 0 at and they did not stand out to me. Do you think it be easier to remove these empty area COMIDS before they are joins with PRMS in the xwalk target @lekoenig ?

```
   PRMS_segid   comid AREASQKM TOTDASQKM
1       119_1 2614050        0    1.5381
2      1265_1 8077738        0  150.4944
3       133_1 2615272        0    8.1063
4       133_1 2615284        0  113.5287
5       158_1 2743164        0 1523.1987
6       158_1 2743508        0 1407.1968
7       158_1 2743520        0 1545.8652
8       158_1 2743522        0 1545.8652
9         2_1 1748739        0   57.1311
10      204_1 4148042        0  240.5691
11     2123_1 9481980        0   36.6264
12     2123_1 9482048        0    1.7253
13     2129_1 9481952        0   82.2807
14      223_1 4150536        0  826.1262
15      226_1 4151818        0  170.7660
16       23_1 1750633        0  607.2930
17      244_1 4153726        0  340.0875
18      249_1 4186295        0  218.6073
19      269_1 4188023        0   11.0610
20      269_1 4188041        0   40.9932
21      275_1 4186425        0  770.5233
22     2765_1 9485884        0   13.1094
23      287_1 4188275        0  127.6074
24      355_1 4490988        0   24.8841
25      365_1 4489216        0    3.8124
26      368_1 4492442        0   11.6073
27      368_1 4492448        0    2.5425
28      374_1 4496436        0   34.5420
29      374_1 4496460        0    3.7152
30      380_1 4495872        0   88.2360
31      380_1 4495884        0   76.8672
32      392_1 4496410        0    8.9190
33      394_1 4496232        0  126.9279
34      395_1 4495910        0   44.7462
35      396_1 4496424        0   94.5765
36       40_1 2585051        0    1.5633
37       45_1 2587011        0    9.9558
38       51_1 2588371        0  728.1810
39      571_1 4648766        0   52.0812
40      579_1 4648744        0   83.4201
41      595_1 4652082        0    6.4584
42      614_1 4655446        0  417.9438
43      878_1 4782271        0  559.7046
44      888_1 4782507        0  181.7802
45      892_1 4782567        0   60.4404
46      901_1 4783437        0  341.9217
47      903_1 4783475        0  174.1113

```

lekoenig:
Good questions. Regarding the specific COMID you mention here, the stream pairing looks correct to me (see plot below) so I don't think this COMID is erroneous. NHDPlusV2 and PRMS are models or representations of the river network, and so the precise spatial location may not always be exact (I also wouldn't expect the basemaps to necessarily reflect the true location of streams). 

nhd_reach

I haven't looked through the NHDPlusV2 user guide to see if it contains further documentation about these reaches where `AREASQKM = 0`, however, I have noticed this before. My guess is that this is due to the area being smaller than the scale of the DEM used to derive the catchments. I like the completeness of the xwalk target in that every COMID associated with a given PRMS segment is stored there, so personally, I'd opt to deal with the reaches where `AREASQKM = 0` in these downstream targets as opposed to filtering them out of the xwalk target. Just my two cents though!

2_process.R
```diff
@@ -61,5 +62,34 @@ p2_targets_list <- list(
...
+               bind_rows()
+            }
+  ),
+  ## Filter LC data to the AOI : DRB and join with COMIDs area info and PRMS ids
+  # returns a df with unique comids for aoi + area of comid and NLCD LC percentage attributes
+  tar_target(p2_LC_w_catchment_area,
+             AOI_LC_w_area(area_att = p1_nhd_area_att,
+                       NLCD_LC_df = p1_NLCD_data$NLCD_LandCover_2011,
+                       aoi_comids_df = p2_drb_comids)
+             ),
```
lekoenig, line 87: This target subsets the NLCD2011 land cover data to only include those COMIDs that map onto the PRMS network in the DRB. Thinking about your comment above about combining `p2_drb_comids` with the `read_subset_LC_data` function, I'm wondering whether it would make sense to merge these (`p1_NLCD_data` and `p2_LC_w_catchment_area`) since we're essentially subsetting/filtering the LC data twice (once for all DRB-AOI COMIDs and once for DRB-PRMS COMIDs)? If we wanted to process the other downloaded LC data (besides NLCD2011) would we need a separate target?

2_process.R
```diff
@@ -61,5 +62,34 @@ p2_targets_list <- list(
...
+  
+  ## Estimate LC proportion in PRMS catchment
+  # returns df with percent LC in PRMS catchment in our AOI
```
lekoenig, line 90: I know it initially took me a minute to realize that the processed land cover attributes were given in proportions rather than percentages. To save ourselves any confusion down the line, I'd suggest we try to keep the naming consistent across column headers, comments, and targets. We're calculating proportions here, is that right?

2_process/src/combine_NLCD_PRMS.R
```diff
@@ -0,0 +1,64 @@
...
+
+###----
+# Estimate proportion of LC coverage at PRMS scale
+proportion_lc_by_prms <- function(NLCD_LC_df_w_area,
+                          catchment_att = "CAT"){
+
+#' @description This function takes the consolidated df from AOI_LC_w_area() and calculated the percentage share of each LC class at the PRMS segid scale
+#' @param NLCD_LC_df_w_area df output of AOI_LC_w_area that has LC data per COMID & PRMS id and area in (km2)
+#' @param catchment_att Specifies the catchment attribute to subset to. Must be either 'CAT', "ACC", 'TOT'. Default = "CAT"
+#' @example proportion_lc_by_prms(NLCD_LC_df_w_area = p2_LC_w_catchment_area, catchment_att = "CAT")
```
lekoenig, line 35: I like the idea of allowing the user to specify "CAT", "ACC", or "TOT", but I'm wondering whether the "ACC" or "TOT" options are meaningful here. For example, if I specified "TOT" I would still sum only the direct catchment areas (given by `AREASQKM`) but then would be multiplying that area by proportion of land cover in the total upstream watershed (`TOTDASQKM`). For total upstream land cover I think we'll need a function that handles land cover differently (joins based on `comid_down` in the xwalk table and then directly choose either "ACC" vs "TOT" attribute values). That could be a separate PR, however, I suggest we omit the upstream choices from this function and rename the function/target to indicate that we're calculating catchment-scale attributes. What do you think? Do you agree with my interpretation above?

2_process/src/combine_NLCD_PRMS.R
```diff
@@ -0,0 +1,64 @@
...
+      across(starts_with(paste0('PRMS_AREA_',catchment_att)), ~(.x/AREASQKM_PRMS),
+             .names="PERCENT_{col}"),
+      # round the new cols 
+      across(starts_with(c('PERCENT','PRMS')), round, 2), 
+    .groups="drop")
+  
+  ## Final cleaning - removing LC categories and PRMS_segis that are empty and the empty PRMS (287_1) 
+  area_df <- area_df %>%
+    filter(AREASQKM_PRMS > 0) %>%
+    select(-contains('NLCD11_12'))
```
lekoenig, line 58: I see how COMIDs where `AREASQKM=0` cause an issue for at least one PRMS segment here, segid "287_1". It makes sense to handle those cases, but filtering that segment out of the data will leave us with a data frame 458 rows long (instead of 459) and I wonder if we'll ever find that confusing when working with these data down the line. Instead of filtering out this row, what if we used `across` to conditionally replace all the land cover values with NA if `AREASQKM_PRMS=0`? Maybe translating this [SO post](https://stackoverflow.com/questions/62578575/conditionally-replace-values-across-multiple-columns-based-on-string-match-in-a) would look like:

`mutate(across(c(PERCENT_AREA_CAT_first_col:PERCENT_AREA_CAT_last_col), ~ifelse(AREASQKM_PRMS==0, NA, .)))`

Do you think this makes sense, or do you think it's better to just exclude "287_1"?

lekoenig (changes requested):
Thanks, @msleckman! These processed NLCD attributes are looking good. I've requested a few minor changes and also pose several general comments/questions that we can discuss prior to merging.

2_process/src/combine_NLCD_PRMS.R
jds485: If we do not have attributes for a reach, the reach would have to be excluded from modeling. So using NA for those reaches is fine. But we know that `AERASQKM` must be greater than 0 for every reach, so I'm wondering if we need to handle these cases before computing the PRMS land cover values. Could you summarize the lengths of the comids with `AREASQKM = 0`? Maybe (length of 0 area comid) / (length of PRMS segment). I am hoping the 0 area segments are all short relative to the length of the PRMS segment, so excluding them from the land cover calculation will have a minor impact.

lekoenig: > And for the 0 area PRMS segment, we could drop it from modeling. Do you know how many SC samples were assigned to that reach?

Right now, it looks like there are no SC samples that are assigned to PRMS segid "287_1":
```
> tar_load(p2_sites_w_segs)
> p2_sites_w_segs %>% filter(subsegid == "287_1")
# A tibble: 0 x 14
# ... with 14 variables: site_id <chr>, site_name <chr>, count_days_nwis <dbl>, count_days_discrete <dbl>, count_days_total <dbl>, lon <dbl>,
#   lat <dbl>, datum <chr>, org_id <chr>, data_src_combined <chr>, subsegid <chr>, offset <dbl>, segidnat <int>, geometry <GEOMETRY [°]>
```

lekoenig: Here's a quick attempt at estimating the proportional length represented by the comids where `AREASQKM=0` -

```
tar_load(p2_prms_nhdv2_xwalk)
tar_load(p1_nhdv2reaches_sf)
tar_load(p1_reaches_sf)

drb_comids <- p2_prms_nhdv2_xwalk %>% 
  tidyr::separate_rows(comid_all,sep=";") %>%
  mutate(AREASQKM=NA, length_km = NA, prms_length_km = NA)

for(i in seq_along(drb_comids$PRMS_segid)){
  nhd_line <- filter(p1_nhdv2reaches_sf,COMID == drb_comids$comid_all[i])
  prms_line <- filter(p1_reaches_sf,subsegid == drb_comids$PRMS_segid[i])
  drb_comids$AREASQKM[i] <- nhd_line$AREASQKM
  drb_comids$length_km[i] <- as.numeric(st_length(nhd_line))/1000
  drb_comids$prms_length_km[i] <- as.numeric(st_length(prms_line))/1000
}

drb_comids %>% 
  mutate(comid_prop_len = length_km/prms_length_km) %>%
  filter(AREASQKM==0) %>% 
  filter(comid_prop_len > 0.05)
```

This returns 5 instances (out of the 47 `AREASQKM=0` comids that Margaux identified) where the proportional length is greater than 5%:
```
# A tibble: 5 x 7
  PRMS_segid comid_down comid_all AREASQKM length_km prms_length_km comid_prop_len
  <chr>           <int> <chr>        <dbl>     <dbl>          <dbl>          <dbl>
1 226_1         4151820 4151818          0   0.00580         0.109          0.0532
2 287_1         4188275 4188275          0   0.0480          0.0479         1.00  
3 51_1          2588373 2588371          0   0.0123          0.105          0.117 
4 878_1         4782271 4782271          0   0.0432          0.0620         0.696 
5 892_1         4782565 4782567          0   0.0143          0.0471         0.303 
```

So I wouldn't expect these comids to have a big impact on the calculated land cover proportions. It looks like there are a few SC observations that are snapped to these PRMS segments:
```
> p2_sites_w_segs %>% filter(subsegid %in% c("226_1","287_1","51_1","878_1","892_1"))
# A tibble: 4 x 14
  site_id      site_name         count_days_nwis count_days_discr~ count_days_total   lon   lat datum org_id  data_src_combin~ subsegid  offset segidnat
  <chr>        <chr>                       <dbl>             <dbl>            <dbl> <dbl> <dbl> <chr> <chr>   <chr>            <chr>      <dbl>    <int>
1 11NPSWRD_WQ~ VALLEY CREEK DS ~              NA                 3                3 -75.5  40.1 NAD83 11NPSW~ Harmonized_WQP_~ 892_1    4.01e-4     2326
2 01444800     Delaware River n~              NA               147              147 -75.1  40.8 NAD83 USGS-PA Harmonized_WQP_~ 51_1     1.15e-3       NA
3 21PA_WQX-WQ~ DELAWARE RIVER                 NA                86               86 -75.1  40.8 NAD83 21PA_W~ Harmonized_WQP_~ 51_1     1.15e-3       NA
4 31DRBCSP-DR~ Delaware River a~              NA                 7                7 -75.1  40.8 NAD83 31DRBC~ Harmonized_WQP_~ 51_1     9.54e-4       NA
# ... with 1 more variable: geometry <LINESTRING [°]>
```

jds485: Thanks for checking that. I'm glad it's just a few PRMS reaches that are affected by this problem. 

I just realized that dropping the PRMS reach whose area is 0 may not be a good plan because it is used in the network connection tables and is an interior segment (as opposed to being on the edge or being a headwater segment). It's also used in the PRMS and streamflow prediction models that will provide input features for this model. So it's better if we can use this reach in the model. Maybe we can approximate the area for just this reach as `length_km^2`?

1_fetch.R
```diff
@@ -94,6 +94,13 @@ p1_targets_list <- list(
     p1_nhdv2reaches_sf,
     get_nhdv2_flowlines(drb_huc8s)),  
 
+  # Extract COMID and respective area for LC estimation below
+  tar_target(p1_nhd_area_att, 
+             p1_nhdv2reaches_sf %>%
+               st_drop_geometry() %>%
+               select(COMID,AREASQKM,TOTDASQKM)
+  ),
+
```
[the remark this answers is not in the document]
msleckman: I agree with this simplification. I removing target `p1_nhd_area_att` in `1_fetch.R` and changed the following target in `2_process.R`  that used `p1_nhd_area_att` as an input: 
```
tar_target(p1_NLCD_data,
             read_subset_LC_data(LC_data_folder_path = p1_NLCD_data_unzipped, 
                                 Comids_in_AOI_df = p1_nhdv2reaches_sf %>% st_drop_geometry() %>% select(COMID), 
                                 Comid_col = 'COMID')
             )
```

1_fetch.R
```diff
@@ -112,11 +119,11 @@ p1_targets_list <- list(
   
   ## read in NLCD datasets and subet by comid in DRB
   ## Note that this returns a vector of dfs if more than one NLCD data is in the p1_NLCD_data_unzipped
+  
   tar_target(p1_NLCD_data,
              read_subset_LC_data(LC_data_folder_path = p1_NLCD_data_unzipped, 
                                  Comids_in_AOI_df = p1_nhdv2reaches_sf %>% st_drop_geometry() %>% select(COMID),
```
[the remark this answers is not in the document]
msleckman: opted to remove p1_nhd_area_att 🤷 😄

2_process.R
msleckman: > Hmm, this target reminds me of your question of when should we put raw code in the target object itself versus a separate function. I know I wrote a version of this code but now as a reader it seems hard to quickly interpret the 2_process file 😸. Maybe we could simplify this target by replacing the split-lapply code with the line below. Does that work (i.e. give the same output)?

> p2_prms_nhdv2_xwalk %>% tidyr::separate_rows(comid_all,sep=";")

Nice job finding this function! Exactly what I am looking for, had not yet found this simple one-liner for this tidy process.  🥳

2_process/src/combine_NLCD_PRMS.R
msleckman: Agreed. Removed this param

2_process.R
msleckman: > I could see the value of using this target as input to read_subset_LC_data. Currently that function takes all the COMIDs within the whole area of interest (~approximately the bbox of the DRB), which leaves us with more COMIDs (and thus more LC values) than we really need. However, the xwalk target that p2_drb_comids stems from gets generated within the 2_process phase, whereas the LC data is initially subset in the 1_fetch phase. Were you thinking of moving one or both targets to combine with the read_subset_LC_data function?

Yea, these are exactly the questions I was thinking about. It requires some decision on what we include in our pipeline's fetch and process. Computationally, its not adding too much to conduct two subsets. Also, I initially saw these two subsets as different: the *1st* in fetch subsets the raw LC data that originally covers all of Conus, so its easier to work with; the *2nd* is to join comid area attributes and to ensure the scaling to PRMS is possible.

msleckman (commented):
@lekoenig finished first process of your comments (push  below)
- Items I've left atm: changes to how we are ingesting NLCD in `1_fetch.R` so that we can combine `1_fetch: read_subset_LC_data()` and `2_process:AOI_LC_w_area()`.  See comments.
- Handling comids with 0 Area.

2_process.R
msleckman: `If we wanted to process the other downloaded LC data (besides NLCD2011) would we need a separate target?`

The specific LC datasets that are read in through this function are dataframes that have a comid key, because they've been built for compilation with selected nhdplus attributes. So this function can be used for any NLCD nhd land cover datasets for other years besides 2011. When processing other LC data besides NLCD nhd, we'll be working with raw tif files and in that case, I think the target should be separate. 


Other key question is whether we'd ever want this LC data at the scale of the DRB-AOI comids (`n=23645`) or if we can immediately subset to DRB-PRMS comids (`n=3229`)

jds485: > Other key question is whether we'd ever want this LC data at the scale of the DRB-AOI comids (`n=23645`) or if we can immediately subset to DRB-PRMS comids (`n=3229`)

I'm not expecting to use the NHD resolution for modeling in this FY, but its something we could explore in later FYs depending on our PRMS network results. Could the NHD data be processed to PRMS with the crosswalk?

msleckman:
@lekoenig is this awaiting any other step? We are sticking with the `p2_prms_nhdv2_xwalk` target and removing the Area = 0, correct?

lekoenig:
@msleckman This is just awaiting merge of the updated xwalk target. The column names in that target have changed slightly, and we'll want to use `comid_cat` to process most VarsOfInterest, including land use proportions here. 

I'd also have to review our conversation from above, but my personal preference is that we don't remove the one segment where Area =0. I think that would confuse me down the line if we only had 458 segments instead of the expected 459. My vote would be for us to handle those segments differently and return NA for the aggregated value (or something similar). Happy to discuss though!

jds485:
I'd also prefer not removing the segment with Area = 0. Going back to my previous comment, can we approximate its area? Which reach is it?

lekoenig:
> I'd also prefer not removing the segment with Area = 0. Going back to my previous comment, can we approximate its area? Which reach is it?

@jds485 I believe this is subsegid == 287_1. Even in the PRMS network it's a short reach so there's only one COMID that intersects and contributes to that segment. Within the NHDPlusV2 attribute tables, `AREASQKM` = 0 for that COMID. There's also no NHDPlusV2 catchment spatial feature that can be downloaded for that COMID, so I'm not sure how to estimate its area other than assigning a small value. In looking at my comment [above](https://github.com/USGS-R/drb-inland-salinity-ml/pull/38#discussion_r778279683) it doesn't look like there are any samples snapped to this segment.

jds485:
> There's also no NHDPlusV2 catchment spatial feature that can be downloaded for that COMID

Does this mean it wouldn't have NHD attributes? 

> I'm not sure how to estimate its area other than assigning a small value

Maybe something like getting neighboring NHD catchment polygons and dissolving them. Assuming there is a hole corresponding to this reach's catchment, QGIS could be used to estimate the gap area by making a polygon that closely matches the gap. It also wouldn't surprise me if there were a function to estimate hole areas in polygons (Arc method [here](https://support.esri.com/en/technical-article/000014007#:~:text=%20How%20To%3A%20Measure%20the%20area%20of%20donut,the%20Catalog%20window%2C%20navigate%20to%20System...%20More%20)).

The reach is on the interior of the network, so it unfortunately can't just be removed from modeling

1_fetch.R
```diff
@@ -107,7 +107,11 @@ p1_targets_list <- list(
         suppressWarnings()
       }
   ),
-
+  
+  tar_target(
+    p1_catchments_sf_valid, st_buffer(p1_catchments_sf,0)
+  ),
+
```
msleckman, line 114: Added `p1_catchments_sf_valid` target

2_process.R
```diff
@@ -48,10 +49,30 @@ p2_targets_list <- list(
   # Pair PRMS segments with intersecting NHDPlusV2 reaches and contributing NHDPlusV2 catchments
   tar_target(
     p2_prms_nhdv2_xwalk,
-    create_GFv1_NHDv2_xwalk(prms_lines = p1_reaches_sf,nhd_lines = p1_nhdv2reaches_sf,prms_hrus = p1_catchments_sf,
+    create_GFv1_NHDv2_xwalk(prms_lines = p1_reaches_sf,nhd_lines = p1_nhdv2reaches_sf,prms_hrus = p1_catchments_sf_valid,
                             min_area_overlap = 0.5,drb_segs_spatial = drb_segs_spatial)
   ),
   
+  ## Melt PRMS_nhdv2_xwalk to get all cols of comids Ids and PRMS ids filtered to drb 
+  tar_target(p2_drb_comids_all_tribs,
```
msleckman, line 57: renamed target

msleckman:
This PR is ready to merge. The items from previous PR have been integrated here. 
I've also added a target to fix geometry of catchments sf object  ([`1_fetch.R`: line 111 - 114](https://github.com/USGS-R/drb-inland-salinity-ml/pull/38#discussion_r794101443)) 
Renamed target `p2_drb_comids_all_tribs`

2_process.R
```diff
@@ -48,10 +49,31 @@ p2_targets_list <- list(
...
+    create_GFv1_NHDv2_xwalk(prms_lines = p1_reaches_sf,nhd_lines = p1_nhdv2reaches_sf,
+                            prms_hrus = p1_catchments_sf_valid,
                             min_area_overlap = 0.5,drb_segs_spatial = drb_segs_spatial)
   ),
   
+  ## Melt PRMS_nhdv2_xwalk to get all cols of comids Ids and PRMS ids filtered to drb 
+  tar_target(p2_drb_comids_all_tribs, 
+            p2_prms_nhdv2_xwalk %>%
+              tidyr::separate_rows(comid_seg,sep=";") %>% 
+              rename(comid = comid_seg)
```
lekoenig, line 61: For processing the land cover proportions (and any area-weighted variable from the MasterList), we'll want to use `comid_cat` here rather than `comid_seg` since `comid_seg` won't include all of the tributaries that contribute to the PRMS HRU. Here's a map for PRMS segid 112_1 that hopefully helps show the differences between the three columns in the xwalk table:
example_segs

2_process/src/combine_NLCD_PRMS.R
```diff
@@ -0,0 +1,62 @@
...
+      across(starts_with(paste0('PRMS_AREA_','CAT')), ~(.x/AREASQKM_PRMS),
+             .names="PROP_{col}"),
+      # round the new cols 
+      across(starts_with(c('PROP','PRMS')), round, 2), 
+    .groups="drop")
+  
+  ## Final cleaning - removing LC categories and PRMS_segid that are empty and the empty PRMS (287_1) 
+  area_df <- area_df %>%
+    filter(AREASQKM_PRMS > 0) %>%
+    select(-contains('NLCD11_12')) %>% select(-starts_with('PRMS_AREA_CAT'))
```
lekoenig, line 56: It seems like the consensus is that we do not want to omit segments without area (which would leave 458 segments in the DRB instead of the expected 459). I looked quickly and the COMIDs where `AREASQKM == 0` all appear to be very short (e.g. < 0.2 km long according to the NHDV2 attributes). In that case, I think we could go with Jared's [suggestion](https://github.com/USGS-R/drb-inland-salinity-ml/pull/38#discussion_r778301198) to approximate area for those reaches as `LENGTHKM^2`. The resulting areas would be very small but would have a value and so we would not need to omit any reaches. 

Would adding a line to the code chunk that generates `area_df` work here? E.g.:

```
area_df <- NLCD_LC_df_w_area %>%
    ## If area = 0, approximate area as length^2:
    mutate(AREASQKM_approx = case_when(AREASQKM ==0 ~ LENGTHKM^2,
                                       TRUE ~ AREASQKM)) %>% 
    # multiply COMID area by the LC percent for all the cols that start with 'CAT" - unit km2
    mutate(across(starts_with('CAT'),~(.x/100)*AREASQKM_approx,.names="AREA_{col}")) %>%
    # aggregate to PRMS scale through group_by()
    group_by(PRMS_segid) %>%
    # summarize group by at PRMS catchment scale
    summarize(
      # Sum to calculate total area of the PRMS  catchment - unit km2
      across(AREASQKM_approx, sum, .names='AREASQKM_PRMS'),
      # Sum to calculate  LC area at the PRMS catchment scale per LC class
      across(starts_with(paste0('AREA_','CAT')), sum, .names = 'PRMS_{col}'),
      # calculate proportion coverage of specified LC class in the entire PRMS catchment - proportion
      across(starts_with(paste0('PRMS_AREA_','CAT')), ~(.x/AREASQKM_PRMS),
             .names="PROP_{col}"),
      # round the new cols 
      across(starts_with(c('PROP','PRMS')), round, 2), 
    .groups="drop")

```

lekoenig, line 56: Now that there is an issue/PR devoted to combining and/or renaming some of the land use categories across datasets, I wonder if omitting `NLCD11_12` should be done within that step of the pipeline? I'm not very familiar with what that step will look like though, so I leave that up to you.

lekoenig (changes requested):
Thanks, @msleckman! I agree that this is just about ready to merge. There are just 2 comments that should be addressed first. These aren't new comments but just reflect changes that we discussed in the comments.

lekoenig:
And to respond to your question above:

> Subsets to the drb comids and joins to the with PRMS segment ids. Note still have ACC and TOT which are filtered out next.
@lekoenig do we need all three comid cols (comid == comid_seg from target p2_prms_nhdv2_xwalk) or can I remove to make the table cleaner?

I don't think we need all three comid cols. It looks like you've already made that change and only retain whatever comid's get returned in `p2_drb_comids_all_tribs` 👍

2_process/src/combine_NLCD_PRMS.R
msleckman: Yea this is a good idea. I'll remove this section the inverse select()

2_process.R
lekoenig: I'm realizing it's not clear from this map, but the segments shown in green are not represented by the PRMS lines (which overlaps the red/blue).

2_process/src/combine_NLCD_PRMS.R
msleckman: @jds485  Circling back to this.

Why length_km^2? 
This approximation does put the value into the area units, however assumes that the catchment is a square around the  reach length, where the length of reach (LENGTHKM) is a side. At least that how I am seeing this.
How about assuming the cross-section of the catchment polygon as `LENGTHKM`. So equiv. to a diagonal of square shape. 
 
 in that case, the new approximated area `= 1/2 length_km^2`

2_process.R
msleckman: that makes sense. Where's the PRMS seg 112_1 ?  Is that overlapping as well?

2_process/src/combine_NLCD_PRMS.R
jds485: Reach length squared does not force the catchment to be shaped like a square, but I understand your reasoning here. I'm not sure which would be a better estimator of true catchment area. We could test it out with the shorter reaches in our dataset to be formal (e.g., by mean square error)

2_process.R
lekoenig: PRMS seg 112_1 is just hidden under the red and blue segments in the map above.

2_process/src/combine_NLCD_PRMS.R
msleckman: I did not run any mean square error testing, but I've evaluated the rougher options we have to estimate the missing area. 

The difference between running  `1/2 length_km^2` and just `length_km^2` is relatively negligible in any case. 
When I apply both to get an approx area for those comids with area =0, the sums differ by 1 km2. 
```
> ## With Approximated area calc as LENGTHKM^2
> area_df_1 <-  p2_LC_w_catchment_area %>%
+   mutate(AREASQKM_approx = case_when(AREASQKM == 0 ~ LENGTHKM^2, TRUE  ~ AREASQKM))
> 
> ## With Approximated area calc as 1/2*LENGTHKM^2
> area_df_2 <-  p2_LC_w_catchment_area %>%
+   mutate(AREASQKM_approx = case_when(AREASQKM == 0 ~ 1/2*(LENGTHKM^2), TRUE  ~ AREASQKM))
> 
> sum(area_df_1$AREASQKM_approx[area_df_d$AREASQKM == 0])
[1] 1.97574
> sum(area_df_2$AREASQKM_approx[area_df_d$AREASQKM == 0])
[1] 0.98787
```
Per comid - we evidently see that `AREASQKM_approx` calc. in `area_df_2` is half that of `area_df_1`

A potential alternative could be to apply to the comid with `area == 0` the mean `AREASQKM` of comids with similar `LENGTHKM`. So, apply the mean `AREASQKM` value of comids with `AREASQKM > 0` & `LENGTHKM` ~= to `LENGTHKM` of comids `area == 0`. 

```
> #alternative 
> ## Takea average lengthkm of comid with 0 area
> avg_length_0_area <- mean(p2_LC_w_catchment_area$LENGTHKM[p2_LC_w_catchment_area$AREASQKM == 0])
> print(avg_length_0_area)
[1] 0.04240077
> 
> # extract area of other comids that are ~ the same length
> mean(p2_LC_w_catchment_area$AREASQKM[round(p2_LC_w_catchment_area$LENGTHKM,2) == round(avg_length_0_area,2)])
[1] 0.0309473
```

Now, ultimately it is prms `287_1` that turning to NA comids are aggregated to PRMS level, because that segid is made up of only one comid (`4188275` ).  For the other PRMS with comids with `area == 0`,  the `AREASQKM` of that PRMS is estimated by the other valid comid areas that it comprises. 
The `LENGTHKM` of PRMS `287_1` is `0.048`. Could we take the average area of all catchments in our basin with a LENGTHKM of 0.048 ? 

```
> len_287_1 <-  p2_LC_w_catchment_area$LENGTHKM[p2_LC_w_catchment_area$PRMS_segid == '287_1']
> p2_LC_w_catchment_area$AREASQKM[p2_LC_w_catchment_area$LENGTHKM == len_287_1]
 [1] 0.0045 0.0018 0.0018 0.0000 0.0000 0.0369 0.0018 0.0027 0.0072 0.0099 0.0000 0.0000 0.0009 0.0099 0.0018 2.1951 0.0018 0.0063 0.0090 0.0036 0.0036 0.0018 0.0000 0.0027 0.0207 0.0054
> mean(p2_LC_w_catchment_area$AREASQKM[p2_LC_w_catchment_area$LENGTHKM == len_287_1])
[1] 0.08958462

```

While I've outlined 3 options here, I left our initial option  (LENGTHKM^2) in the code for now.

2_process/src/combine_NLCD_PRMS.R
msleckman: added to Issue #71

2_process/src/combine_NLCD_PRMS.R
lekoenig: I'm in favor of either 1) assuming area ~ length^2 for any COMID with area = 0, or 2) assuming area ~ some DRB mean for very short reaches.

I'm tempted to go with option 1 just for simplicity. As far as I can tell, the longest NHD reach with zero area is 193 meters. The reach that is equivalent to PRMS segid 287_1 is 48 meters long. So these are really small and I'm guessing our decision here will have a negligible impact on the variable estimates at the PRMS-scale. Even for 287_1, for instance, there are no SC samples that currently get snapped to that segment, so I think assigning an area is less about getting an accurate estimate and more about having a placeholder value so that the segment gets retained in this data processing step. What do you think, @jds485?

2_process/src/combine_NLCD_PRMS.R
```diff
@@ -0,0 +1,58 @@
...
+      # Sum to calculate total area of the PRMS  catchment - unit km2
+      across(AREASQKM_approx, sum, .names='AREASQKM_PRMS'),
+      # Sum to calculate  LC area at the PRMS catchment scale per LC class
+      across(starts_with(paste0('AREA_','CAT')), sum, .names = 'PRMS_{col}'),
+      # calculate proportion coverage of specified LC class in the entire PRMS catchment - proportion
+      across(starts_with(paste0('PRMS_AREA_','CAT')), ~(.x/AREASQKM_PRMS),
+             .names="PROP_{col}"),
+      # round the new cols 
+      across(starts_with(c('PROP','PRMS')), round, 2), 
+    .groups="drop")
```
lekoenig, line 52: Do you think we'll want areas of each land use type? In your data cleaning step before, you had a step ` %>% select(-starts_with('PRMS_AREA_CAT'))` that is now omitted.

lekoenig (approved):

lekoenig:
Thanks, Margaux! I approve of these changes and the pipeline builds for me - you can merge whenever you're ready 😃

2_process/src/combine_NLCD_PRMS.R
msleckman: Ohh right. Didn't remember i removed that. good find! ... Yes I removed that because I figured we'd keep just the proportion values and we weren't interested in the area values. I'll put back this col filter code statement because it makes the output smaller and significantly easier to work with.

lekoenig: OK, sounds good to me since there are lots of columns with the land use data. I wasn't sure if you had decided otherwise, so just figured I'd double-check. Thanks for your reply : )

2_process.R
```diff
@@ -48,14 +49,37 @@ p2_targets_list <- list(
...
+                            prms_hrus = p1_catchments_sf_valid,
                             min_area_overlap = 0.5,drb_segs_spatial = drb_segs_spatial)
   ),
   
+  ## Melt PRMS_nhdv2_xwalk to get all cols of comids Ids and PRMS ids filtered to drb 
+  tar_target(p2_drb_comids_all_tribs, 
+            p2_prms_nhdv2_xwalk %>%
+              tidyr::separate_rows(comid_cat,sep=";") %>% 
+              rename(comid = comid_cat)
+  ),
```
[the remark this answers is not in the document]
msleckman: 👍  sounds good! So I am guessing this targets `p2_drb_comids_all_tribs` for now at least.

Issue #30, opened 2021-12-13
Check gauges for tidal influence to inform study reaches


jds485:
Tidal influence can extend up to 30 river km upstream of an estuary. We should check that our sites do not show this signal, or have a weak signal.

For reaches downstream of Trenton, NJ, check daily and instantaneous data within 30 km of the estuary for a tidal signal. I think that would appear in a simple Fourier frequency analysis.

lekoenig:
I agree that some sort of spectral analysis would be a good way to check for tidal influence in the SC data. I started out by looking at a [site](https://waterdata.usgs.gov/nj/nwis/uv/?site_no=01412350&PARAmeter_cd=00010,00095,00300,00301,00400,63680) which we've previously omitted because we have strong reason to think it _is_ tidal - it's in Delaware Bay. I used this site as a control of sorts, to see what kind of signal we'd be looking for.

**Control site 01412350**

1. Download the data and pad the time steps
```
ts_data <- dataRetrieval::readNWISuv(
  siteNumbers = "01412350",parameterCd="00095",startDate = "1979-10-01",endDate = "2021-12-13",tz = "UTC") %>%
  dataRetrieval::renameNWISColumns() %>%
  rename("Value_Inst" = "Lower_SpecCond_Inst")

# Pad timestamps (we need continuous data for spectral analysis):
  ts_data_complete <- ts_data %>%
  tidyr::complete(dateTime = seq(lubridate::ymd_hms(min(dateTime)),lubridate::ymd_hms(max(dateTime)), by = '6 mins')) %>%
  # interpolate values for up to 10 timesteps (in this case, 1 hr):
  mutate(Value_Inst = zoo::na.approx(Value_Inst,maxgap=10))
```
2. For each time series, find the longest continuous stretch of non-NA values on which to perform spectral analysis
```
find_longest_cont_subset <- function(data){
  
  run <- rle(!is.na(data$Value_Inst))
  data$run <- rep(run$lengths,run$lengths)
  data$cond <- rep(run$values,run$lengths) 
  data_cont_subset <- data %>% 
    filter(cond==TRUE) %>%
    filter(run == max(run))
  
}

ts_data_subset <- find_longest_cont_subset(ts_data_complete)
```

3. Plot the power spectrum computed using discrete time fast fourier transform

```
plot_power_spectra <- function(data_subset,delta_t = 1/96,plot_title){
  
  # Calculate the frequency spectrum using fast fourier transform, fft:
[...]

PR #47, opened 2022-01-10
Check sites for tidal signal

lekoenig:
The code changes in this PR create a new target `p2_sites_w_segs_nontidal_csv` that represents an updated site list. We are omitting discrete samples in reaches we suspect are tidally-influenced (see discussion in issue #30), so this target filters out discrete sites that are snapped to the mainstem Delaware River below Trenton, NJ. 

mainstem
 
This PR also makes the following changes:

- edits to the sitelist_summary log file to reflect the omitted discrete sites in `p2_sites_w_segs_nontidal_csv`
- I've updated `dummy_date` in `_targets.R` to re-build the NWIS data pipeline. Namely, I'd like to reset our expectations for the log files since there was some confusion around these files with recent commits and at least one of the log files was missing from `main`. As a consequence of this change, the pipeline may take awhile to re-build locally.


_targets.R
```diff
@@ -49,14 +49,19 @@ site_tp_select <- c("ST","ST-CA","SP")
 # sites 01412350, 01484272, and 01482695 coded as site type "ST" but appear to be tidally-influenced
 omit_nwis_sites <- c("01412350","01484272","01482695") 
 
+# Define PRMS mainstem segments thought to be tidally-influenced (used for filtering sites used in daily/sub-daily models)
+mainstem_reaches_tidal <- c("2771_1","2769_1","2768_1","2767_1","2764_1","2762_1","2763_1","2759_1","2757_1","2752_1",
+                            "2753_1","2755_1","2772_1","388_1","389_1","385_1","386_1","383_1","382_1","377_1","378_1",
+                            "376_1","351_1","344_1","346_1","333_1")
+
```
jds485, line 56: This is fine for now, but we could also use the reach network connectivity table to select the reaches downstream of the mainstem Trenton reach. `reach_attributes_drb.csv` from [ScienceBase](https://www.sciencebase.gov/catalog/item/5f6a289982ce38aaa2449135). I can make an issue to create targets for the datasets in that ScienceBase item because we'll use them for modeling and add to the reach attribute tables for this work

jds485:
Thanks for letting me know it may take a while to rebuild. I'll start running now and plan to finish the review this afternoon

2_process/src/create_site_list.R
```diff
@@ -140,3 +140,31 @@ transform_site_locations <- function(site_list_df,crs_out){
...
+  #' 
+  
+  site_list_filtered <- site_list_w_segs_df %>%
+    # For daily-resolution model, omit discrete sites associated with reaches thought to be tidal:
+    filter(!(subsegid %in% mainstem_segs & data_src_combined=="Harmonized_WQP_data")) %>%
+    # For discrete sites that are associated with tidal reaches and are also NWIS sites, exclude discrete samples from count_days tally:
+    rowwise() %>%
+    mutate(count_days_discrete = case_when((subsegid %in% mainstem_segs) ~ 0, 
+                                           TRUE ~ count_days_discrete),
+           count_days_total = sum(count_days_nwis,count_days_discrete,na.rm=TRUE)) %>%
```
jds485, line 160: Are the NWIS and discrete days guaranteed to be mutually exclusive? Or is this double counting days that have both NWIS and discrete?

lekoenig: These days are not guaranteed to be mutually exclusive and I would expect some overlap (double-counting of days) that will be reconciled in the merged time series.

2_process/src/create_site_list.R
```diff
@@ -140,3 +140,31 @@ transform_site_locations <- function(site_list_df,crs_out){
...
+  #' @description Function to filter sites thought to be influenced by tides 
+  #'
+  #' @param site_list_df data frame containing site locations and matched segment id's
+  #' @param mainstem_segs character string containing the mainstem segments that are assumed to be influenced by tides
+  #' @param fileout file path and name for output data, including the file extension
+  #' 
+  
+  site_list_filtered <- site_list_w_segs_df %>%
+    # For daily-resolution model, omit discrete sites associated with reaches thought to be tidal:
+    filter(!(subsegid %in% mainstem_segs & data_src_combined=="Harmonized_WQP_data")) %>%
```
jds485, line 155: This filter removes discrete sites, but there are some discrete samples at NWIS sites that we do not want to include. The count of discrete sampling days is updated to be correct for those NWIS sites, but where are are the discrete samples removed somewhere?

jds485: I'd prefer adding to this PR. Would it then be possible to update the `count_days_discrete` from the adjusted target instead of setting to 0 here? Or, maybe better, you could write a filtered site list from the sites in the adjusted target

lekoenig: Yep, I could do that! I honestly kind of wavered back and forth between the two approaches so let's go with the version you suggest here.

_targets.R
lekoenig: OK, good idea. We've downloaded those reach attributes within the DO pipeline, so we could easily port here as well (e.g. https://github.com/USGS-R/drb-do-ml/pull/37).

jds485 (changes requested):
Changes are in the comments

_targets.R
lekoenig: I've added issue #53 to create targets for the network attributes, including the network connectivity matrix.

lekoenig:
Here's an outline of the changes I made based on PR reviews:

- add `quiet` argument to `st_read` calls to suppress messages from sf (issue #49)
- remove `print()` statement in get_nlcd_LC.R (issue #49)
- made edits to `summary_multiple_inst_ts.csv` to make log files more consistent among local builds (issue #52)
- subset the discrete (WQP) conductance data to only include samples from nontidal sites
- from the newly-filtered WQP dataset, create a filtered site list that reflects site numbers and observation-day tallies for nontidal sites only (although note that for our initial modeling purposes, this code retains NWIS continuous sites - "uv" or "dv" - associated with tidal reaches; see discussion in issue #47 )

1_fetch/src/find_sites_multipleTS.R
```diff
@@ -0,0 +1,31 @@
...
+#' @param omit_sites character string indicating any NWIS site numbers to exclude. Default is NULL.
+#' @param fileout file path and name for saved data table, including the file extension
+#'
+#' @examples 
+#' find_sites_multipleTS(site_list = p1_nwis_sites,earliest_date = "1980-10-01",latest_date = "2020-09-30",fileout="1_fetch/summary_multiple_inst_ts.csv")
+#'
+
+multTS_table <- site_list %>%
+  # retain "uv" sites that contain data records after user-specified {earliest_date}
+  filter(data_type_cd=="uv",!(site_no %in% omit_sites),end_date > earliest_date) %>%
```
jds485, line 18: >= for earliest date? I'd expect the earliest date to be included

jds485 (approved):
Looks good! I made 2 comments with small changes

1_fetch/src/find_sites_multipleTS.R
jds485: Okay, thanks for that explanation. Checking for more than one day of data could be a useful quality check for sites, so I'm okay with keeping as `>`

lekoenig:
Thanks for your comments, @jds485 - they were really helpful for improving the code in this PR.

Issue #58, opened 2022-01-20
Add container files to repo and connect to the USGS container registry


jds485:
Files to add:
- Dockerfile
- docker-compose.yml 

Link with GitLab container registry

jds485:
Closing because this is now in our GitLab repo

PR #62, opened 2022-01-21
Add Docker compatibility

jds485:
Added Dockerfile and re-ran the pipeline within the resulting container. This PR will address Issue #58, but is not yet complete.


jds485:
I think this can be merged into main now so that we can all use the container. I added the instructions from the tutorial as a README file. Jeff: I know you have WSL already, so could you try building the repo from the instructions? There are a couple of steps to set up WSL2 with Docker Community Edition. I ran the “Configuring WSL” section of this repo by Jeremy Fee to set up Docker-CE https://code.chs.usgs.gov/jmfee/wsl-environment. Let me know if you have any trouble!

jsadler2:
@jds485 - sorry. this slipped through the cracks. I will give this a shot.

docker/README.md
```diff
@@ -0,0 +1,54 @@
...
+Windows filesystem:
+
+`docker build -t drb-insal-gs:v0.1 /mnt/c/users/jsmith/OneDrive\ -\ DOI/Documents/Projects/DRB-InlandSalinity/drb-inland-salinity-ml/docker/`
+
+Linux filesystem: I cloned our repo to /home/jsmith/
+
+`docker build -t drb-insal-gs:v0.1 /home/jsmith/drb-inland-salinity-ml/docker/`
+
+
+## Running the container and testing the build of our repo
```
jsadler2, line 25:
```suggestion
## Running Rstudio in the container
```

docker/README.md
```diff
@@ -0,0 +1,54 @@
...
+- -v mounts your files to the server
+- --rm removes the container after the session ends
+- -e sets your username
+- your password will print to the screen
+- go to localhost:8787 in a browser to access your Rstudio session
+
```
jsadler2, line 31: I recommend putting these lines after the docker commands (the commands that currently start in line 34). That way people see the commands and then if they are curious to know what the different flags are they can learn more below.

jsadler2: I also think wording it like this would be a little clearer (you can edit the wording in step 4 :) )
```suggestion
Explanation of options:
- -v mounts your files to the server
- --rm removes the container after the session ends
- -e sets your username

Once the server has started:
1. your password will print on the screen
2. go to localhost:8787 in a browser
3. put in your username (from the `-e` option above) and password (generated by the server)
4. make some magic with Rstudio in your browser

```

docker/README.md
```diff
@@ -0,0 +1,54 @@
+# Instructions for building the Docker container from the Dockerfile and running code within the container
+
+## Starting docker daemon within WSL2 or Linux OS
+`sudo service docker start`
+
+`sudo dockerd > /dev/null 2>&1 &`
+
+
+## Building an image from the Dockerfile
```
jsadler2, line 9: Most of the time we shouldn't have to do a `docker build` right? Once the image is built and pushed to the container registry, we should just be able to do a `docker pull.` I think it'd be worth saying that here. What do you think?

jsadler2 (changes requested):
This looks good, @jds485. I was able to build the image and run the pipeline (until the xwalk target) in the container. I have few minor comments.

docker/README.md
jds485: Yes, that's right. I'll add a comment about that here

docker/Dockerfile
```diff
@@ -0,0 +1,65 @@
+#Starting with the lastest stable geospatial R image as of 1/20/2022
+#Dockerfile references are here:
+#https://github.com/rocker-org/rocker-versioned2/tree/master/dockerfiles
+#includes Ubuntu 20.04 LTS (stable release)
+#includes R version 4.1.2
+#includes all packages and Linux libraries in the Dockerfile for verse, 
+#which is based on tidyverse, which is based on rstudio, which is based on r-ver.
+FROM rocker/geospatial:4.1.2
+
```
jesse-ross, line 9: It may be worthwhile to add the public key for the DOI's SSL Intercept. This helps the system work while users are connected to VPN. Cf. [here](https://code.usgs.gov/wma/wp/reanalysis-gefs/-/blob/main/docker/Dockerfile#L8-9) for an example.

docker/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+  rgdal \
+  rgeos \
+  sbtools \
+  scico \
+  sf \
+  targets \
+  tarchetypes \
+  terra \
+  tidyverse \
+  visNetwork
```
jesse-ross, line 40: `install2.r` leaves some tarballs in `/tmp/` while will make this image a little bit bigger than it needs to be. You can tack an `rm` on to the end of this command to make it a little slimmer, e.g. [here](https://code.usgs.gov/wma/wp/reanalysis-gefs/-/blob/main/docker/Dockerfile#L29).

jesse-ross (commented):
I stumbled on this and figured I'd add a couple quick notes on the `Dockerfile`. I love the thorough and clear comments!

docker/Dockerfile
jds485: Thanks for the review! I'll add these changes

jds485:
I added a [container registry repo](https://code.usgs.gov/wma/wp/drb-inland-salinity-container/container_registry) for us on GitLab and pushed the latest version of the container to that repo. I'm not sure if there's a way to do a Merge Request for a container registry.

I'm wondering how is best to connect this code repo and the container repo, and if we need to connect them at all. I'm leaning towards not needing to connect them. 
- The Dockerfile would be stored and tracked in the GitLab repo, and we can update it as needed to run the code repo. 
- In the code repo, we could track the container version corresponding to each target in a log file. 
  - The date targets are built is listed in the `tar_meta()$time`, so I think that could be used to grab the targets built in the most recent tar_make run. We'd record the start time at the top of `_targets.R` and use that to search the meta file for all targets built after that time. The resulting log file would track the target name and the container name (for now). Jake gave examples of other details we could include.

Do you think it would be better for me to move this docker PR to the GitLab repo?

I would appreciate your thoughts @jesse-ross

jesse-ross:
> I'm wondering how is best to connect this code repo and the container repo, and if we need to connect them at all. I'm leaning towards not needing to connect them.

I think there's no need to connect them. You can treat code.usgs.gov just like Docker Hub if you want, like you've done. But I do like your idea of moving the Dockerfile over to the repo you set up for the container registry, because then you could build with CI (e.g. [here](https://code.usgs.gov/wma/wp/reanalysis-gefs/-/blob/main/.gitlab-ci.yml)). The advantages would be
1. you can easily and consistently insert a lot of metadata about how the container was built into the image itself
2. you can tag the image with a reference to the git commit from which it was built. This gives unique identifiers which won't get overwritten, as things like "v0.1" and "latest" tend to have happen to them.

jds485:
@jsadler2 I think this PR can be closed after you approve the GitLab docker PR. I copied the files in this PR to GitLab and modified them. The _targets.R file in this branch has a few new lines related to S3 that are better to move to another PR that is dedicated to S3.

Issue #45, opened 2022-01-06
Add tributary catchments to NHD-PRMS crosswalk 


lekoenig:
In reviewing PR #38 I wanted to check that the summed areas for the contributing NHD catchments were approximately equal to the area of the PRMS catchment polygons associated with each reach. This check shows that for many catchments, the PRMS catchment area is much greater than our summed NHD area calculated from the comid's in the crosswalk target:

![catchment_addition](https://user-images.githubusercontent.com/8785034/148428586-90a3213a-97cc-47c3-b0ad-e5a8253a2e5e.png)

This is because our crosswalk was focused on finding those NHD reaches that intersect the PRMS segments, which excludes the headwater tributaries (and their catchments) that are not included in the PRMS network: 

![headwater_problem](https://user-images.githubusercontent.com/8785034/148428763-e9067144-1d92-4c9d-8c4e-02aa84875561.png)

Therefore, in order to collate land use information as in PR #38 , we will need to add a column to the crosswalk target that includes these headwater catchments that contribute to the PRMS-scale catchment.

lekoenig:
I've been working on a fix to gather all of the NHD catchments that would directly drain to a given PRMS segment (as described above, this isn't correct in the current xwalk table for larger PRMS catchments that are missing contributing tributary COMID's). 

Here's my current solution:

1. Each PRMS segment has a unique value of `comid_down`, the NHD COMID at the bottom of that segment. The GF also refers to these NHD reaches as points of interest or POIs. For each `comid_down`, find all upstream COMIDs.
2. Find other POIs in the xwalk table that are located upstream of the target PRMS reach. For each of those, find all upstream COMIDs
3. By difference, identify the COMIDs that are returned in the subset of reaches from 1) but not located in 2). These are the COMIDs that intersect the local PRMS catchment.

I've attempted to display those steps for the example catchment above here:

![fix](https://user-images.githubusercontent.com/8785034/149383035-0b0fb56d-d55e-4603-a071-994bfb33e60e.png)

In general, this really improves our PRMS-area to summed-NHD-area relationship: 

![compare_areas](https://user-images.githubusercontent.com/8785034/149383168-be290029-0089-4161-9a3a-11af25228c98.png)

[...]

PR #63, opened 2022-01-24
Pair NHDPlusV2 catchments to PRMS segments

lekoenig:
This PR makes edits to the prms-nhdv2 crosswalk target such that the target now contains a column called `comid_cat` that represents all of the COMIDs that drain directly to each PRMS segment ID in the DRB. The format of this column is the same as before. Additional details and discussion are in issue #45.

### DIFF
--- a/2_process.R
+++ b/2_process.R
@@ -3,6 +3,7 @@ source("2_process/src/munge_inst_timeseries.R")
 source("2_process/src/create_site_list.R")
 source("2_process/src/match_sites_reaches.R")
 source("2_process/src/pair_nhd_reaches.R")
+source("2_process/src/pair_nhd_catchments.R")
 
 p2_targets_list <- list(
   
@@ -43,15 +44,23 @@ p2_targets_list <- list(
      p2_sites_w_segs,
      get_site_flowlines(p1_reaches_sf, p2_site_list, sites_crs = 4269, max_matches = 1, search_radius = 0.1)),
   
-  # Pair PRMS segments with NHDPlusV2 reaches
+  # Pair PRMS segments with intersecting NHDPlusV2 reaches and contributing NHDPlusV2 catchments
   tar_target(
     p2_prms_nhdv2_xwalk,
     {
-      p1_reaches_sf %>%
+      # find NHDPlusV2 COMID's that intersect PRMS segments
+      reach_to_seg_xwalk <- p1_reaches_sf %>%
         split(.,.$subsegid) %>%
         purrr::map(.,pair_nhd_reaches,nhd_lines=p1_nhdv2reaches_sf) %>%
         purrr::map(.,summarize_paired_comids) %>%
         bind_rows()
+      
+      # find NHDPlusV2 COMID's that drain directly to PRMS segments
+      p1_reaches_sf %>%
+        split(.,.$subsegid) %>%
+        purrr::map(.,pair_nhd_catchments,prms_hrus=p1_catchments_sf,min_area_overlap=0.5,nhd_lines=p1_nhdv2reaches_sf,xwalk_table = reach_to_seg_xwalk) %>%
+        bind_rows() %>%
+        left_join(reach_to_seg_xwalk,.,by="PRMS_segid")
     }
   ),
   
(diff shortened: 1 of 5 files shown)


2_process/src/pair_nhd_catchments.R
```diff
@@ -0,0 +1,124 @@
...
+  #' 
+    
+  # subset xwalk for prms_line
+  xwalk_prms_line <- xwalk_table %>%
+    filter(PRMS_segid == prms_line$subsegid)
+  
+  # Find all tributaries upstream of comid_down
+  nhd_reach_down_UT <- nhdplusTools::get_UT(nhd_lines,xwalk_prms_line$comid_down)
+  
+  # Find GF POI's (=respective comid_down's) that are upstream of the PRMS segment of interest
```
jds485, line 105: do you define GF elsewhere? Geospatial Fabric

2_process.R
```diff
@@ -43,15 +44,23 @@ p2_targets_list <- list(
...
         purrr::map(.,summarize_paired_comids) %>%
         bind_rows()
+      
+      # find NHDPlusV2 COMID's that drain directly to PRMS segments
+      p1_reaches_sf %>%
+        split(.,.$subsegid) %>%
+        purrr::map(.,pair_nhd_catchments,prms_hrus=p1_catchments_sf,min_area_overlap=0.5,nhd_lines=p1_nhdv2reaches_sf,xwalk_table = reach_to_seg_xwalk) %>%
+        bind_rows() %>%
+        left_join(reach_to_seg_xwalk,.,by="PRMS_segid")
     }
```
jds485, line 64: Might help to make this a separate function because it's long

jds485: It may also be worth implementing targets parallelization for this. It took ~20 mins to complete on my laptop. I can file parallelization as an issue for the repo.

2_process/src/pair_nhd_catchments.R
```diff
@@ -0,0 +1,124 @@
...
+  #' 
+  #' @examples pair_nhd_catchments(prms_line = p1_reaches_sf[p1_reaches_sf$subsegid == "3_1",],
+  #'                               prms_hrus = p1_catchments_sf,
+  #'                               min_area_overlap = 0.7,
+  #'                               xwalk_table = p2_prms_nhdv2_xwalk,
+  #'                               nhd_lines = p1_nhdv2reaches_sf)
+  #'                               
+  
+  # Identify PRMS segments that require special handling
+  drb_segs_spatial <- c("31_1","135_1","233_1","236_1","244_1","249_1","332_1","354_1","355_1","358_1","396_1","593_1","1256_1","2137_1",
```
jds485, line 26: this could be declared in _targets.R and as an argument to the function. I don't think we'll need to change this list, but it could be helpful to have as an option

2_process/src/pair_nhd_catchments.R
lekoenig: Maybe not (I guess I just leaned into the USGS-speak there). I can make that change in the revision to this PR, though.

2_process.R
lekoenig: I agree that this might be too long for the phase file. I was on the fence about creating a separate function and decided against it here so that you can see the functions that are used to create the target (`pair_nhd_reaches`,`summarize_paired_comids`, and `pair_nhd_catchments`). I could definitely make that change though.

jds485: Sounds good! Don't worry about the parallelization. That's better for a different PR

lekoenig: I didn't realize targets had options for parallelization but it'd be great to speed things up when possible (the crosswalk target takes ~15 min to build on my machine). Thanks for creating an issue - what do you think about implementing those changes in a separate PR that is just focused on parallelization/optimization changes?

2_process/src/pair_nhd_catchments.R
lekoenig: OK, that sounds good. I placed it within the function since that vector shouldn't change, but an advantage of moving to the _targets.R file is that it's more visible.

2_process.R
jds485: I opened up an issue for parallelization. One thing I/we need to check with the Data Science Branch is best practices for parallelizing targets that download from NWIS. I think Lindsay said they try not to do that, but it's just a vague memory for me. Do you know? I made a similar comment in the issue, so we can continue the discussion there

jds485 (approved):
Looks good! I think this is ready to merge

Issue #43, opened 2022-01-04
Gather monthly natural baseflow data for DRB


lekoenig:
Download monthly estimates of natural baseflow referenced to NHDPlusV2 reaches from ScienceBase (target 1980-2015): https://www.sciencebase.gov/catalog/item/6023e628d34e31ed20c874e4

jds485:
Because this dataset does not cover 2015-present, I was thinking its value for modeling depends on how much interannual variability there is for each month. Here are 2 possible uses:

1. We could treat it as a static feature instead of a dynamic one. To make it static, monthly averages could be computed over the 35 years. 
2. Alternatively, we could use it as a dynamic monthly feature and fill in the 2016-present data with the monthly averages from the last 35 years.

lekoenig:
The monthly estimates of natural baseflow are referenced to NHDPlusV2 reaches within the DRB. There are a few decision points for processing:

**1. NHD-to-PRMS translation**
Currently, I grab the values from the COMID's that represent the downstream ends of each PRMS segment. In doing so I assume that the monthly average baseflow for that COMID applies across the extent of the PRMS reach. However, some PRMS segments are comprised of multiple NHDPlusV2 reaches, so alternatively we could also take an average of the COMID-level baseflow estimates for all COMIDs that intersect a given PRMS segment.  

**2. Summary stats of interest**
This dataset contains four summary statistics for each month/year, including the mean and median predicted monthly baseflow from a random forest ensemble of 1000 predictions as well as the 10th and 90th percentiles of those predictions. I currently focus on the mean values but we could also retain all four values. Could the median/10th percentile/90th percentile predictions of monthly baseflow be useful as additional features?

**3. Static vs dynamic monthly feature**
In the interest of time, I suggest we forego running the random forest model to predict values for 2016-present, at least for now. Therefore, we could either compute monthly averages for the last 35 years, or use the baseflow estimates as a dynamic feature.

> We could treat it as a static feature instead of a dynamic one. To make it static, monthly averages could be computed over the 35 years.

It looks like there's a decent amount of interannual variability in the (mean predictions of) monthly natural baseflow. Here's an example for one 5th order reach that intersects our PRMS network:

![interannual_variation](https://user-images.githubusercontent.com/8785034/151071543-18becc20-a0f1-4fbd-87f4-38a6a9799c1b.png)

[...]

PR #65, opened 2022-01-27
Fetch natural baseflow data

lekoenig:
This PR fetches natural baseflow estimates for the DRB from ScienceBase (Miller et al. 2021). The monthly natural baseflow estimates are referenced to NHDPlusV2 COMIDs so the code changes here use the crosswalk table to:

- identify the COMID at the downstream end of each PRMS segment, 
- return the monthly estimates for those COMIDs, and 
- munge the data into a single output table containing monthly natural baseflow estimates from 1979-2021 for each PRMS segment 
 
Some additional discussion of processing steps is in issue #43.

### DIFF
--- a/1_fetch.R
+++ b/1_fetch.R
@@ -5,6 +5,7 @@ source("1_fetch/src/find_sites_multipleTS.R")
 source('1_fetch/src/get_nlcd_LC.R')
 source("1_fetch/src/get_nhdplusv2.R")
 source("1_fetch/src/get_gf.R")
+source("1_fetch/src/fetch_sb_data.R")
 
 p1_targets_list <- list(
   
@@ -145,6 +146,16 @@ p1_targets_list <- list(
     p1_vars_of_interest,
     read_csv(p1_vars_of_interest_csv, show_col_types = FALSE)
   ),
+  
+  # Download monthly natural baseflow for the DRB
+  # from Miller et al. 2021: https://www.sciencebase.gov/catalog/item/6023e628d34e31ed20c874e4
+  tar_target(
+    p1_natural_baseflow_zip,
+    download_sb_file(sb_id = "6023e628d34e31ed20c874e4",
+                     file_name = "baseflow_partial_model_pred_XX.zip",
+                     out_dir="1_fetch/out"),
+    format = "file"
+  )
 
 )
   
--- a/1_fetch.R
+++ b/1_fetch.R
@@ -155,6 +155,16 @@ p1_targets_list <- list(
                      file_name = "baseflow_partial_model_pred_XX.zip",
                      out_dir="1_fetch/out"),
     format = "file"
+  ),
+  
+  # Unzip monthly natural baseflow file
+  tar_target(
+    p1_natural_baseflow_csv,
+    {
+      unzip(zipfile=p1_natural_baseflow_zip,exdir = dirname(p1_natural_baseflow_zip),overwrite=TRUE)
+      file.path(dirname(p1_natural_baseflow_zip), list.files(path = dirname(p1_natural_baseflow_zip),pattern = "*.csv"))
+      },
+    format = "file"
   )
 
 )
--- a/1_fetch.R
+++ b/1_fetch.R
@@ -118,10 +118,11 @@ p1_targets_list <- list(
   
   ## Unzip all NLCD downloaded datasets 
   ## Note - this returns a string or vector of strings of data path to unzipped datasets 
-  tar_target(p1_NLCD_data_unzipped, 
-             unzip_NHD_NLCD_data(downloaded_data_folder_path = p1_NLCD_data_zipped,
-                                 create_unzip_subfolder = T),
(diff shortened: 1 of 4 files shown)


1_fetch.R
```diff
@@ -118,10 +119,11 @@ p1_targets_list <- list(
   
   ## Unzip all NLCD downloaded datasets 
   ## Note - this returns a string or vector of strings of data path to unzipped datasets
```
jds485, line 121: thanks for doing the formatting changes!

2_process.R
```diff
@@ -62,6 +63,17 @@ p2_targets_list <- list(
...
-    format = "file")
+    format = "file"),
+  
+  # Return natural baseflow estimates for each PRMS segment
+  tar_target(
+    p2_natural_baseflow,
+    munge_natural_baseflow(baseflow_pred_files = p1_natural_baseflow_csv,
+                           segs_w_comids = p2_prms_nhdv2_xwalk %>% 
+                             select(PRMS_segid,comid_down) %>% 
+                             rename('COMID' = 'comid_down'),
```
jds485, line 74: Do you expect that we'll need this renamed column for other functions?

2_process.R
```diff
@@ -62,6 +63,17 @@ p2_targets_list <- list(
...
+                           start_year = as.character(lubridate::year(earliest_date)),end_year = "2021")
```
jds485, line 75: I think the end would always correspond to the dummy_date. Do you agree? `end_year = as.character(lubridate::year(dummy_date))`

2_process/src/munge_natural_baseflow.R
```diff
@@ -0,0 +1,81 @@
...
+  monthly_baseflow_ls <- lapply(baseflow_pred_files,function(x){
+    
+    # Extract the month for each monthly baseflow file
+    # this line finds numbers with 2 digits within the file name string
+    month <- str_extract(x, "[0-9]{2}")
+    
+    # Read in monthly tables
+    data <- read_csv(x,show_col_types = FALSE) %>%
+      # reformat column names to include month
+      rename_with(.cols = p10.Pred:mean.Pred, function(x){paste0(x,"_", month)}) %>%
```
jds485, line 27: I've never tried making a function of x inside of a function of x. I guess the scopes are independent and it works, but I'd prefer changing to a different variable here (y?)

jds485:
Looks good! I'm just running it now. For some reason the `p2_prms_nhdv2_xwalk` is rebuilding as well. I didn't see a change made to that in this PR, though

2_process/src/munge_natural_baseflow.R
```diff
@@ -0,0 +1,81 @@
+munge_natural_baseflow <- function(baseflow_pred_files,segs_w_comids,vars=c("mean","med","p10","p90"),start_year,end_year){
```
jds485, line 1: Could you add a logical argument `fill_all_years` that allows for using only monthly baseflow to fill all years? That would involve adding an if statement around `years_to_fill`

lekoenig:
> For some reason the p2_prms_nhdv2_xwalk is rebuilding as well. I didn't see a change made to that in this PR, though

Huh, I wouldn't have expected that either. The natural baseflow target does rely on the crosswalk table, but aside from some formatting changes to the phase files I can't think of any upstream changes that would initiate a rebuild.

jds485 (approved):
Looks good! Just some minor changes in the comments. You can merge after addressing those

2_process.R
lekoenig: I don't really expect to use the column labeled 'COMID' again. In the crosswalk table I like keeping the column name `comid_down` because it helps signify that that column represents the COMID at the downstream end of a given PRMS segment. 

I renamed the column here rather than using `comid_down` in the function to 1) make a join step within the function _slightly_ easier to read/interpret, and 2) to perhaps give a user more latitude in choosing the COMIDs of interest beyond having a specific column called `comid_down`. If the renaming step makes the target more difficult to read, I'd opt for just requiring a column named `comid_down` for the function `munge_natural_baseflow`.

2_process/src/munge_natural_baseflow.R
lekoenig: Ooh, good catch! It does seem to work but I agree with you that replacing the second function call to depend on y would be easier to read and probably also safer.

2_process/src/munge_natural_baseflow.R
lekoenig: Good idea, I'll add that argument!

2_process.R
jds485: Okay, thanks. I think it's easier the way you have it; I was just curious if the name change should be permanent

lekoenig:
Thanks, @jds485! I've made the changes you suggested, including adding an argument to `munge_natural_baseflow` to indicate whether missing years should be filled.

Issue #72, opened 2022-02-01
Fix ScienceBase links in NHD attribute database


jds485:
This issue is to identify the broken links and add the new links to the [NHD attribute spreadsheet](https://github.com/USGS-R/drb-inland-salinity-ml/blob/main/1_fetch/in/NHDVarsOfInterest.csv)

jds485:
@lekoenig mentioned that the mean monthly precipitation dataset is the only broken link. That is addressed in #73, so I'm linking that PR to this issue

PR #73, opened 2022-02-01
Edited mean monthly precip dataset info

jds485:
Addresses #72.

I changed the mean monthly precipitation dataset to 1945-2015. These are annual datasets, so I've entered them into the spreadsheet with YYYY as a placeholder for the year. That matches the format used for other variables.


1_fetch/in/NHDVarsOfInterest.csv
```diff
@@ -9,18 +8,18 @@ Climate,FSTFZ6190,"Watershed average of mean day of the year of first freeze, de
...
+Climate,PPT_OCTYYYY,50 year (1945 - 2015) catchment mean of October precipitation (mm) based on Wolock and McCabe (2018),CAT_PPT_OCTYYYY,ACC_PPT_OCTYYYY,TOT_PPT_OCTYYYY,millimeters,https://www.sciencebase.gov/catalog/item/5730cf1ce4b0dae0d5db1f17,NA,NA,FALSE,area_weighted_mean
+Climate,PPT_NOVYYYY,50 year (1945 - 2015) catchment mean of November precipitation (mm) based on Wolock and McCabe (2018),CAT_PPT_NOVYYYY,ACC_PPT_NOVYYYY,TOT_PPT_NOVYYYY,millimeters,https://www.sciencebase.gov/catalog/item/5730cf1ce4b0dae0d5db1f17,NA,NA,FALSE,area_weighted_mean
+Climate,PPT_DECYYYY,50 year (1945 - 2015) catchment mean of December precipitation (mm) based on Wolock and McCabe (2018),CAT_PPT_DECYYYY,ACC_PPT_DECYYYY,TOT_PPT_DECYYYY,millimeters,https://www.sciencebase.gov/catalog/item/5730cf1ce4b0dae0d5db1f17,NA,NA,FALSE,area_weighted_mean
 Climate,PPT7100_ANN,"Mean annual precip (mm) for the watershed, from 800m PRISM data.  30 years period of record 1971-2000. ",CAT_PPT7100_ANN,ACC_PPT7100_ANN,TOT_PPT7100_ANN,mm/year,https://www.sciencebase.gov/catalog/item/573b70a7e4b0dae0d5e3ae85,NA,NA,FALSE,area_weighted_mean
```
lekoenig, line 23: Should we still gather the 30-year mean annual precip derived from PRISM?

lekoenig (approved):
Thanks, @jds485 - looks good!

1_fetch/in/NHDVarsOfInterest.csv
jds485: yes. Because the 1945-2015 data are provided annually, I think it will be helpful to only use the 1971-2000 years to match the annual timeframe and the timeframe for all other meteorological variables. That detail probably should be noted in the processing notes for that dataset. Do you have a suggestion for where to add this note?

lekoenig: I've created an issue (#80) to track special handling tasks like using only a subset of dates here for `PPT7100_ANN`. For now, I'd say we should add a brief note in an in-line comment in the code. There's also a section within InlandGoal3-DataProgress.docx called 'Climate Data.' That could be a good spot to jot down notes related to variable processing.

jds485: Thanks! I think the other datasets are good to be used as-is after the crosswalk and aggregation method listed in the spreadsheet column, but that's based on the written description. I'll let Margaux know there's an issue for tracking additional processing steps for these attributes

lekoenig: OK, thanks - I've noticed some flagged values (e.g. -9999) so I might add that to issue #80, too.

jds485: Oh, good catch! And good idea to add to 80. I saw those in the land cover PRs and I suggested changing to NA values. Does that sound good?

lekoenig: Sounds reasonable to me!

Issue #53, opened 2022-01-11
Fetch DRB network attributes from ScienceBase


lekoenig:
Download network attributes from ScienceBase: https://www.sciencebase.gov/catalog/item/5f6a289982ce38aaa2449135

PR #75, opened 2022-02-02
Fetch DRB network attributes

lekoenig:
This PR addresses issue #53. The code changes add three targets to download and read in the reach attributes and network adjacency matrix from ScienceBase (https://www.sciencebase.gov/catalog/item/5f6a289982ce38aaa2449135). Are there any other files within that SB repo that we are interested in (e.g. reservoir_features_drb.csv)?


jds485:
Thanks! Could you add `sntemp_inputs_outputs_drb.csv`? Let's hold off on the reservoir attributes for now. Those can be added if your analysis of SC data near reservoirs or model residuals near reservoirs suggest a different treatment is needed

lekoenig:
> Could you add sntemp_inputs_outputs_drb.csv? Let's hold off on the reservoir attributes for now.

Yep! I've added sntemp_inputs_outputs_drb.zip to the download target, and added two targets to unzip and read in that information.

jds485 (approved):
Looks good and it ran for me. You can merge when ready

PR #89, opened 2022-02-17
Removing crosswalk functions from repo. Reading in crosswalk table

### LINKED ISSUE #86: Move PRMS-NHDv2 xwalk out of salinity repo
Move NHGFv1-NHDv2 xwalk out of salinity repo and instead manually add as an input dataset in 1_fetch/in

jds485:
I removed variables and functions that were only used to build the crosswalk. They are now in [drb-network-prep](https://github.com/USGS-R/drb-network-prep).

My one uncertainty is: Is it okay to remove `drb_segs_spatial`? Those are the PRMS segments that needed special handling in the crosswalk. I'm wondering if we'd want to save those for future steps.

I kept the target name the same with prefix `p2`, even though it's a fetch step. Is that okay, or should it be moved to fetch?

Closes #86


2_process.R
```diff
@@ -57,9 +54,7 @@ p2_targets_list <- list(
   # Pair PRMS segments with intersecting NHDPlusV2 reaches and contributing NHDPlusV2 catchments
   tar_target(
     p2_prms_nhdv2_xwalk,
-    create_GFv1_NHDv2_xwalk(prms_lines = p1_reaches_sf,nhd_lines = p1_nhdv2reaches_sf,
-                            prms_hrus = p1_catchments_sf_valid,
-                            min_area_overlap = 0.5,drb_segs_spatial = drb_segs_spatial)
+    read_csv(GFv1_NHDv2_xwalk_url, col_types = 'cccc')
```
lekoenig, line 57: When I run `tar_make()` I get an error on `p2_natural_baseflow` - that target is relying on a tabular join with key = COMID (where COMID is an integer in the data we grab from SB). Do you get that error too? I'm wondering whether we need to keep `comid_down` as an integer or just specify the COMID column as character within the `p2_natural_baseflow` target (line 27 of 2_process/munge_natural_baseflow.R).

lekoenig (changes requested):
Thanks, @jds485 - I'm happy to see that we can read in the xwalk table from github with the raw file ✨ 

I had one error message come up related to the expected column type of `comid_down` but otherwise the code built for me. Do you mind checking out that comment and letting me know whether your local build also snags at p2_natural_baseflow?

2_process.R
jds485: I'm in the habit of making IDs characters, primarily because they sometimes start with 0 or are long and get truncated when treated as numeric (not sure if that's true for COMIDs). I'm more in favor of converting the `baseflow_pred_files` to a character COMID. I just searched for "comid_down" and didn't see other code to change.

jds485: Thanks! I specified the column types for all columns

lekoenig (approved):
Thanks, @jds485 - these changes look good and the pipeline builds for me. I realized I didn't address your question about `drb_segs_spatial`. I don't foresee needing those in this pipeline so I agree with your decision to remove that object in `_targets.R`. You can merge whenever you're ready.

PR #93, opened 2022-02-18
Implement parallelization option

### LINKED ISSUE #64: Implement parallelization option
Some targets take >20 mins to build. It would help to add tar_future or tar_clustermq parallelization options. 

I think data download targets are not supposed to be run in parallel so that the servers are not overloaded with requests (please correct me if that's wrong). We can use the `deployment = 'main'` option within `tar_targets` to force downloads to happen on the main core. Also open to alternative methods to using targets (e.g., force the download to happen with 1 worker instead of on main)

jds485:
This PR implements clustermq parallelization. 

Code additions:

- Added `deployment = 'main'` to download targets and log file targets. I also used main for targets with very fast processing times (mostly in p1).
- Added line breaks and changed style of some targets in 1_fetch.R and 3_vizualize.R. I think these now match style in other files. 

I tested that our pipeline builds on Tallgrass in the container with multicore parallelization over 8 cores. I only tested p2 and p3 targets build in parallel because most of p1 is run in serial (download targets). The slurm script and the R script that was run are in this PR. It took about 20 mins to build all p2 and p3 targets.

Closes #64


jsadler2 (approved):
Looks good to me, @jds485

PR #81, opened 2022-02-04
Land cover standardization

msleckman:
First Stab at the land cover standardization PR. 

Fine-tuning, testing, and some pruning still to do, but I am at a stage where PR can be submitted. 

**Description:**
This PR primarily adds the `reclassify_land_cover.R` file containing 1 function `reclassify_land_cover()`. 
Functions works with NLCD and the FORESCE input datasets. 
Takes a land cover dataset and reclassifies it's land cover classes to a standardized classification, detailed in a pre-defined reclassification table (current classification has 9 classes, vs 20+ in NLCD). The reclassification tables have been also added to `1_fetch/in/`

**Output example:**

NLCD: 

```
# A tibble: 458 × 11
   PRMS_segid AREASQKM_PRMS prop_NLCD11_1 prop_NLCD11_4 prop_NLCD11_5 prop_NLCD11_6 prop_NLCD11_7 prop_NLCD11_8 prop_NLCD11_2 prop_NLCD11_3 prop_NLCD11_9
   <chr>              <dbl>         <dbl>         <dbl>         <dbl>         <dbl>         <dbl>         <dbl>         <dbl>         <dbl>         <dbl>
 1 1_1                65.0           0             0.03           0            0.85          0             0             0.08          0.02          0.01
 2 10_1                5.47          0.02          0.05           0            0.76          0             0.01          0.09          0.02          0.05
 3 11_1                5.94          0.04          0.03           0.1          0.72          0             0.01          0.09          0             0.01
 4 111_1              41.6           0             0.02           0            0.62          0.01          0             0.28          0.07          0   
 5 112_1              84.1           0             0.05           0            0.6           0             0.01          0.23          0.08          0.02
 6 113_1               3.40          0             0.2            0            0.63          0.02          0.01          0.03          0.07          0.02
 7 114_1              54.5           0             0.04           0            0.74          0.01          0.02          0.15          0.03          0.01
 8 115_1              41.7           0.01          0.07           0            0.56          0             0.01          0.24          0.07          0.04
 9 116_1              41.8           0             0.03           0            0.74          0             0.01          0.2           0.02          0   
10 117_1              12.2           0             0.01           0            0.87          0             0.01          0.1           0.01          0   

```

FORESCE: 

```
# A tibble: 761 × 22
[...]

PR #99, opened 2022-03-01
Simple land cover standardization branch edit

msleckman:
Just need to fix one item that was not pushed and merged in #81 . 

In the code, `rename_with()` won't match if `11` is in the pattern, because this was removed as an argument in `reclassify_land_cover()`. This was made in #81 but not pushed before merging PR to USGS-R/main.


jds485 (approved):
Sorry for not catching this in the last review. You can merge when ready.

PR #100, opened 2022-03-02
Change when log files are updated

### LINKED ISSUE #91: Updating the log file tracking method
As discovered in the [drb-network-prep repo](https://github.com/USGS-R/drb-network-prep/pull/1#discussion_r809269337), log files can be built before the targets that they log. To avoid this, the log files function can be supplied the vector of targets that it logs to force a dependency.

jds485:
Forces log files to update when their dependencies are updated instead of immediately, as discussed in #91.

This is not yet ready for review.

Closes #91


jds485:
Looks like I accidentally branched from my open docker PR branch instead of from my main branch. I'm going to close this PR and reopen with my main branch instead.

PR #104, opened 2022-03-03
Forcing dependencies in log files

### LINKED ISSUE #91: Updating the log file tracking method

jds485:
It seems like our repo only had one log file that did not already have a target dependency: `p3_wqp_ind_csv`. For that target, the previous target name was `p2_wqp_SC_csv`, which was not a target name in our pipeline. I changed it to `p2_wqp_SC_data`.

Closes #91

### DIFF
--- a/3_visualize.R
+++ b/3_visualize.R
@@ -43,7 +43,12 @@ p3_targets_list <- list(
   # Create and save indicator file for WQP data
   tar_target(
     p3_wqp_ind_csv,
-    command = save_target_ind_files("3_visualize/log/wqp_data_ind.csv","p2_wqp_SC_csv"),
+    command = {
+      #forcing dependency to the target because the character string 
+      #of the target name does not enforce it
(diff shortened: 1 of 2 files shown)


lekoenig:
Thanks for checking on this and catching the dependency issue with `p2_wqp_SC_data`. The code changes look good and the pipeline builds (and creates a non-empty ind file for that target), so I think this PR is ready to merge.

As an aside, I think the log files will be more useful once we're using a shared data cache instead of local copies of `tar_meta()`. Here's the outcome when I built the pipeline using your PR branch:

hash

I think this small PR is helpful for illustrating that.

lekoenig (approved):

PR #106, opened 2022-03-03
Add special handling for site 01434498

### LINKED ISSUE #105: Add special column handling of NWIS site
Site 01434498 has odd column names. We have been handling these sites manually in the `get_daily_nwis_data` function.

```
colnames(site_data)
[1] "agency_cd" "site_no" "Date" "Value_Max"
[5] "Value_Max_cd" "..4.._Value_Max" "..4.._Value_Max_cd" "..2.._Value_Max"
[9] "..2.._Value_Max_cd" "..3.._Value_Max" "..3.._Value_Max_cd" "..4.._Value"
[13] "..4.._Value_cd" "..2.._Value" "..2.._Value_cd" "..3.._Value"
[17] "..3.._Value_cd" "Value" "Value_cd"
```

jds485:
@lekoenig How did you know which of the columns corresponded to the main river channel? I used the same columns as you selected for the instantaneous data. I couldn't tell from the web page which columns would be which.

Closes #105

### DIFF
--- a/1_fetch/src/get_daily_nwis_data.R
+++ b/1_fetch/src/get_daily_nwis_data.R
@@ -16,7 +16,7 @@ get_daily_nwis_data <- function(site_info,parameter,stat_cd_select,start_date = 
   #' get_daily_nwis_data(site_info = daily_sites,parameter="00095",stat_cd_select="00003",start_date = "2020-10-01",end_date="2021-09-30")
   
   message(sprintf('Retrieving daily data for %s', site_info$site_no))
-
+  
   # Download daily data
   site_data <- dataRetrieval::readNWISdv(
     siteNumbers = site_info$site_no,parameterCd=parameter,statCd=stat_cd_select,startDate = start_date,endDate = end_date) %>%
@@ -47,6 +47,8 @@ get_daily_nwis_data <- function(site_info,parameter,stat_cd_select,start_date = 
                         # and multiple piezometers. Select data that are representative 
                         # of the main river channel:
                         "01434498" = site_data %>%
+                          mutate(Value = `..4.._Value`,
+                                 Value_cd = `..4.._Value_cd`) %>%
                           select(agency_cd,site_no,Date,Value,Value_cd,Value_Max,Value_Max_cd))
   }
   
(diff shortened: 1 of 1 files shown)


jds485:
As discussed in our meeting, I'll look into the site info to see if the column names specify the main channel, and let you know when this is ready for review

jds485:
To check the column names, I used `readNWISdv()` and `readNWISuv()`

The column names for this NWIS site come in differently for the daily data and the instantaneous data. The daily have the names in a format like: `X_..4.._00095_00003`, and the instantaneous data are more descriptive: `X_Piezometer.3.WQ...Piezometer3._00095_00000`. If I assume the order is the same for the daily and the instantaneous data, then I've pulled the main channel in this PR. @lekoenig, do you know of another way to check to what sensors the columns the daily data correspond?

lekoenig (commented):
The code changes here look good! I agree with your assumption that `Value` (originally `X_00095_00001` before we renamed the cols using `dataRetrieval`) corresponds to the main channel. 

In terms of your question about the column headers, I haven't experienced this problem before where the column headers will be informative for the uv/inst data, but are just numbers for the dv/daily data. I don't think this is a `dataRetrieval` issue, since the NWIS webpage for the daily data shows the same thing (check out the 'Available Parameters' [here](https://waterdata.usgs.gov/ny/nwis/dv/?site_no=01434498&agency_cd=USGS&amp;referred_module=sw)). I wouldn't be surprised if the names get updated by the WSC down the line. If that does happen, I still expect `Value` to work in our pipeline since the main channel will probably keep the name "Specific cond at 25C". 

There's more mysteries lurking here though. 

I plotted the inst and daily data for this site as a way of quickly checking that our assumption to use `Value` was a good one. And the axes are on pretty different scales (_strange.._).

site_4498

Quickly looking at these data, I think there might be data entry issues here. For example, looking at the NWIS graph for the last couple months, the daily maximum values are much lower than the daily min/mean values:

image

I'm not really sure where the min/mean values in that dv data graph are coming from since there are no observations where SC > 100 uS/cm in the inst dataset. So this is kind of strange - I "approve" your PR but not necessarily the new data we ingest as a result. I suggest we merge this PR and create a new issue to check that dv data from this site are not being used in the model-ready time series.

jds485:
Thanks for checking the data details! The min, mean, max lines are strange for every sensor at this site. Looks like the main channel for the mean is X_..4.._00095_00003, and for max is X_00095_00001. I can make that change here, but I expect that these column names will eventually be fixed, so it could be safer to drop this site's dv data for now.

lekoenig:
Good catch on those column assignments, Jared! I've plotted those against my own quick daily summaries and agree that the "..4.." column best corresponds to the main channel mean but that the "Value_Max" column best corresponds to the main channel max. I think it's worth contacting the NWIS contact/local WSC about this, but I'd suggest we merge these changes in the meantime and keep an eye out for updates to the naming conventions at this site.
plots

PR #109, opened 2022-03-09
update handling of columns for site 01434498

### LINKED ISSUE #107: Future: Daily values for site 01434498
Described in #106, the column names for daily values for this site are out of order and not named well. We're expecting that they will eventually be changed. This issue is to remind us that this site has expected future issues. I'm not adding this to the Project yet.

jds485:
Some diagnostics:

Red is for max, black for mean:
image

```
all(test$Value_Max >= test$Value, na.rm = T)
[1] TRUE
```

Closes #107


lekoenig (approved):
This looks good! Thank you, NY WSC 💖 

Updating the original plot I had that showed big differences in SC magnitude between the inst and daily data:
inst

Issue #112, opened 2022-03-17
Reassess sites-to-segs search radius 


lekoenig:
In USGS-R/drb-do-ml#36, we realized that the search radius being used to match sites to segments likely resulted in many headwater streams being matched to the mainstem rivers represented in the PRMS network. This issue includes the following steps:

- [x] Look at a few specific conductance sites to assess whether this issue also applies to our SC sites
- [x] If the search radius should be adjusted, incorporate the code changes in USGS-R/drb-do-ml#67

jds485:
Thanks for discovering this issue! I'm supportive of reducing the search radius to 1 km (or shorter) and dropping sites that are further away than that. It would be helpful to know how many sites, particularly continuous, are dropped to potentially encourage modeling at finer NHD scale in the future.

lekoenig:
We're currently snapping points to segments by searching for the nearest segment within a radius of 0.1 degrees (~10 km). The plot below shows the distance between our SC sites (including discrete + NWIS observations) and the matched segment, showing that there are quite a few sites that are matched to reaches that are relatively far away (i.e., ~1360 sites out of 3449 total sites are > 1 km away from their matched segment). 

![snapped_dist](https://user-images.githubusercontent.com/8785034/159008636-550e5750-aba3-4489-bd5c-f12939e1a9d0.png)

Most of these sites that don't get matched/are matched at long distances are discrete sites. However, there are some NWIS sites that are getting snapped to PRMS segments even though the sites probably aren't located on those segments. NWIS [01478950](https://waterdata.usgs.gov/monitoring-location/01478950/#parameterCode=00065&period=P7D) (on Pike Creek and _not_ the Christina River, which is shown in red) is an example of that:

![snap_ex](https://user-images.githubusercontent.com/8785034/159016439-cd16e43f-ccb5-4df6-a321-09fc5c88c57a.png)

lekoenig:
[...]

PR #116, opened 2022-03-18
Edit search radius used for site-to-reach matching

lekoenig:
Addresses #112.

When matching SC sites to PRMS segments we're currently using a search radius of 0.1 degrees (~10 km), although this is probably too broad and results in headwater tributaries getting improperly snapped to the larger mainstem rivers represented in PRMS (see #112 for further details). 

I've edited `get_site_flowlines()` so that the search_radius is given in meters rather than degrees, and that the offset returned by `nhdplusTools::get_flowline_index()` is also in meters. The following changes are also included: 

- Rename the variable attribute offset to bird_dist_to_subseg_m because the latter is more informative and matches the terminology used in other projects (delaware-model-prep).
- Instead of returning a data frame containing _all_ sites (where `bird_dist_to_subseg_m` is `NA` for unmatched locations), the function now returns a sparse data set that only contains sites that are matched to a segment as determined by the search_radius _or_ are specified by the user as sites to retain.

If we use 500 m here (while adding an argument to retain the two NWIS sites on the Delaware River that are >500 m away from their respective segments), `p2_sites_w_segs` goes from 3,523 rows to 1,845 rows and we lose 9,048 NWIS observation-days ( (~5% of our total NWIS obs-days for SC).

Here's a preview of `p2_sites_w_segs`:

```
> dim(p2_sites_w_segs)
[1] 1845   13
> head(p2_sites_w_segs) %>% as.data.frame()
                     site_id                                  site_name count_days_nwis count_days_discrete count_days_total       lon
1 11NPSWRD_WQX-HOFU_BOYER_01        FRENCH CREEK AT PA ROUTE 345 BRIDGE              NA                   1                1 -75.76927
2 11NPSWRD_WQX-HOFU_BOYER_04   FRENCH CREEK AT COUNTY PARK ROAD (T-452)              NA                   1                1 -75.72541
3 11NPSWRD_WQX-HOFU_BOYER_06  FRENCH CREEK AT SHEEDER MILL ROAD (T-491)              NA                   1                1 -75.62782
4 11NPSWRD_WQX-HOFU_BOYER_07    BIRCH RUN AT HOLLOW AND FRENCH CK ROADS              NA                   1                1 -75.62062
5 11NPSWRD_WQX-HOFU_BOYER_08   FRENCH CREEK AT HARES HILL ROAD (SR1045)              NA                   1                1 -75.56773
6 11NPSWRD_WQX-HOFU_BOYER_10 FRENCH CREEK AT RR TRESTLE IN PHOENIXVILLE              NA                   1                1 -75.51104
       lat datum       org_id   data_src_combined subsegid bird_dist_to_subseg_m segidnat
1 40.20343 NAD83 11NPSWRD_WQX Harmonized_WQP_data    863_1              2.249200     2297
2 40.16987 NAD83 11NPSWRD_WQX Harmonized_WQP_data    863_1             13.987362     2297
3 40.15085 NAD83 11NPSWRD_WQX Harmonized_WQP_data    863_1              7.507442     2297
4 40.14754 NAD83 11NPSWRD_WQX Harmonized_WQP_data    863_1             37.533292     2297
5 40.14096 NAD83 11NPSWRD_WQX Harmonized_WQP_data    888_1             15.870168     2322
6 40.13601 NAD83 11NPSWRD_WQX Harmonized_WQP_data    888_1              1.110031     2322
> 
```

### DIFF
--- a/2_process/src/match_sites_reaches.R
+++ b/2_process/src/match_sites_reaches.R
@@ -2,23 +2,26 @@
 #' from https://code.usgs.gov/wwatkins/national-site-reach/-/blob/master/R/match_sites_reaches.R
 #' modified by: Jeff Sadler
 #' Match each site with a reach (seg_id/COMID)
-get_site_flowlines <- function(reach_sf, sites, sites_crs, max_matches = 1, search_radius = 0.1) {
+get_site_flowlines <- function(reach_sf, sites, sites_crs, max_matches = 1, search_radius = 500, retain_sites = NULL) {
   #' 
   #' @description Function to match reaches (flowlines) to point sites 
   #'
-  #' @param reach_sf sf object of reach polylines with column "subsegid" and in WGS84
-  #' @param sites dataframe with columns "lat" "lon"
-  #' @param sites_crs the crs os the sites table (i.e., 4269 for NAD83)
+  #' @param reach_sf sf object of reach polylines with column "subsegid"
+  #' @param sites dataframe with columns "site_id", "lat", and "lon"
+  #' @param sites_crs the crs of the sites table (i.e., 4269 for NAD83)
   #' @param max_matches the maximum number of segments that a point can match to
-  #' @param search_radius the maximum radius in same units as sf object
-  #' within which a segment will match (segments outside of the radius will not match)
+  #' @param search_radius the maximum radius in meters to match a point to a segment;
+  #' segments outside of search_radius will not match
+  #' @param retain_sites character vector indicating particular sites to retain, even if 
+  #' distance between the site and the nearest flowline segment exceeds the search_radius;
+  #' the site is indicated by the `site_id` column within the sites data frame.
   #'
   #' @value A data frame the same columns as the sites input dataframe with additional columns
-  #' of "segidnat", "subsegid" and "offset" where "offset" is the distance (in degrees) between the point
-  #' and matching flowline
+  #' of "segidnat", "subsegid" and "bird_dist_to_subseg_m" where "bird_dist_to_subseg_m" is the 
+  #' distance (in meters) between the point and matching flowline
   
   # set up NHDPlus fields used by get_flowline_index
-  # Note: that we are renaming the subsegid column to COMID because the nhdplusTools
+  # Note: we are renaming the subsegid column to COMID because the nhdplusTools
   # function requires an sf object with a "COMID" column. This does not have anything
   # to do with the actual nhd COMIDs 
   # Note: the `ToMeas` and `FromMeas` are also required columns for the nhdplusTools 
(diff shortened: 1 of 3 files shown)


2_process/src/match_sites_reaches.R
```diff
@@ -27,36 +30,53 @@ get_site_flowlines <- function(reach_sf, sites, sites_crs, max_matches = 1, sear
...
+  # since we are interested in a more accurate estimate of the `offset` distance between a 
+  # point and the matched reach, set precision to 1 m.
+  # Conduct initial search using a larger radius (search_radius*2) than specified to 
+  # account for any uncertainty in the RANN:nn2 radius search. Then 
+  # filter sites to include those within the specified search_radius.
   flowline_indices <- nhdplusTools::get_flowline_index(flines = reaches_nhd_fields,
                                                        points = sites_sf,
                                                        max_matches = max_matches,
-                                                       search_radius = search_radius) %>%
+                                                       search_radius = search_radius*2,
```
jds485, line 57: Why *2 now? Is this to allow NWIS sites to be matched and later retained? If that's why, I think it would be better to add a separate search radius column for the NWIS sites to retain. This would allow for still retaiing NWIS sites when the `search_radius` is reduced to something smaller than would match the NWIS sites

jds485 (commented):
Looks good, but I haven't run this yet because I have a question for you (see comment).

2_process/src/match_sites_reaches.R
lekoenig: Yeah, good question. I made this change to initially use a wider search radius when we apply `nhdplusTools::get_flowline_index()`. That function uses a radius search implemented in `RANN::nn2()` and I noticed when looking at DO data that a couple sites were not being picked up with a 500 m search radius (but were picked up with a 1000 m radius) even though the distance between those sites was < 500 m, as confirmed by the offset returned and separately using {sf}. More details are [here](https://github.com/USGS-R/drb-do-ml/pull/67#discussion_r827244387). 

So this doesn't have anything to do with the two NWIS sites but rather, to ensure we're picking up all the sites we would expect initially, which we then filter to the requested `search_radius` a couple lines down. I'm open to other suggestions, too, so happy to hear your thoughts!

jds485: Strange that those 2 sites do not get picked up! I'm not familiar with the `nn2` function, and the source is in C, so I'm less inclined to look into if it's a matter of using a different argument in `get_flowline_index`

lekoenig: Yeah, I wasn't worried about those two sites in particular, but the behavior was puzzling to me. The distances returned seem to be correct, so for now I decided to just adjust the argument in `get_flowline_index` to make sure we're capturing all the sites we'd expect. I tried to explain this reasoning in the comments because I know the different argument to search_radius is a little odd.

jds485 (approved):
You can merge when you're ready

Issue #84, opened 2022-02-11
Add S3 with targets 


jds485:
- [x] Git commit the targets metadata
- [x] Add `format =` to all targets and set options as global variables
- [ ] Select which targets to send to S3 with format = aws_... . It's best to commit small targets to GitHub and large targets to S3. 
- [ ] Git commit the targets we do not send to S3. 
- [ ] (nice to have) Set up a flag for global `format` variables that auto-detects if the repo is being run from USGS-R main (commit data to S3) or from a forked copy (save data locally)
- [ ] Set up auto-generation of S3 credentials and pulling of access keys from credential file. These expire in 1 hr by default, so add a variable to increase the default time. Add to .Renviron. Does this need to be added to the container, too?

PR #119, opened 2022-03-18
Add targets to S3

jds485:
In-progress PR to add targets to S3.

Adds the targets metadata to git tracking. Stops ignoring the _targets folder. The _targets folder has its own .gitignore file that I changed to not ignore the meta information.

Global variables `format_file` and `format_aws` are added to the `_targets.R` script to define formats for file and non-file objects. 

Fixes some line indentation in R scripts.

Addresses #84


jsadler2:
superseded by #151

PR #122, opened 2022-03-21
Refine features

### LINKED ISSUE #96: Refine static features for modeling
Based on the exploratory analysis of static features that resulted from #95, some static features require further investigation and down-selection for modeling. Here are some observations from looking into the summary table and map/plots:

- [x] Remove CAT_Physio_Area
- [x] Drop attributes with no spatial variation in value (e.g., all = 0)
- [x] Check several features with NA values in some reaches (13 features total). Can they be filled in from neighboring reaches? Do we drop the feature? Can stream density be calculated from the NHD data (stream length / area)? _*Recharge and water table variables are only NA in the estuary. These make sense as NA for the water table and 0 for recharge.*_
- [x] Olsen permeability pattern is different from SSURGO permeability. Is this rock vs. soil? - yes
- [x] Check that spatial variation is sufficient to be useful and not just act as a trend or local effect.
- [x] Almost surely there is an error in the average water table depth (CAT_EWT_area_wtd) for a reach. -140 m is more than double the 2nd deepest depth, and that is physically very unlikely. Could be that it's in ft. instead of m, or decimal is in the wrong place. Or is this a very small area reach?
- [x] The minimum catchment elevation (CAT_ELEV_MIN_min) seems like it's in the middle of the network and is an order of magnitude larger than the 2nd lowest elevation, which is suspicious. _*This is a quarry, so it's accurate*_

jds485:
This PR refines the static features with the following changes:
- Remove CAT_Physio_Area: this represents the area not covered by a physiographic province.
- Remove RUN7100: this attributes seems to be computed for a coarser model resolution (figure)
CAT_RUN7100_area_wtd

- Drop attributes with no spatial variation in value (e.g., all = 0)
- Fills in CAT Basin Areas that are 0 with PRMS areas (5 total segments)
- Fills in recharge NAs from the average of immediately neighboring reaches. 
- Fills in an extreme outlier water table depth from the average of immediately neighboring reaches
- Calculates stream density (stream length / area) for 5 NA segments from the NHD data in each PRMS catchment
- Computes TOT variables from CAT variables for 4 attributes: consecutive wet days, average annual temperature, average minimum temperature, and stream density. For now, I saving these as separate attributes with `fromCAT` added to the variable name. I think it's safe to overwrite the original values with these values computed from CAT. Example: minimum temperature as reported
TOT_TMIN7100

minimum temperature computed from CAT
TOT_TMIN7100_frmCAT

Updated visualizations are provided for these refined attributes in target `p3_nhdv2_attr_refined_png`.

Remaining to do:

- [x] after review, remove original TOT columns that were re-computed from CAT variables

Nice to have:
- [x] Related to #126: Correctly compute TOT variables from CAT variables for PRMS segments with _1 and _2. Currently, only _2 is computed correctly. I'm thinking this is a "nice to have" because TOT values for _2 vs. _1 should be similar. I would appreciate suggestions for how to handle this. I'll add a comment where this special handling is needed.
- [x] Could add analysis of correlation for CAT and TOT separately, then all together to see if there are any near perfect relations. Are any TOT and CAT variables the same for any attributes?

Closes #96


2_process/src/process_nhdv2_attr.R
```diff
@@ -269,3 +269,184 @@ process_catchment_nhdv2_attr <- function(file_path,vars_table,segs_w_comids,nhd_
...
+  ind_areas <- mutate(ind_areas, str_dens = length_m/CAT_BASIN_AREA_sum) %>%
+    select(-length_m, -CAT_BASIN_AREA_sum)
+  #assign to attribute table
+  nhdv2_attr_refined <- mutate(nhdv2_attr_refined,
+                          CAT_STRM_DENS_area_wtd = case_when(PRMS_segid %in% ind_areas$PRMS_segid ~
+                                                               ind_areas$str_dens[match(PRMS_segid, ind_areas$PRMS_segid)],
+                                                             TRUE ~ CAT_STRM_DENS_area_wtd)
+  )
+  
+  #Compute TOT variables from PRMS CAT variables
```
jds485, line 354: This is where special handling for PRMS segments with _1 and _2 would be needed.

jds485:
I looked into computing TOT from CAT variables for segments with _1 and _2. I started by looking at the areas and was surprised that the TOT area is so much larger for _2 compared to _1 segments. 
```
> p2_nhdv2_attr_refined$TOT_BASIN_AREA[p2_nhdv2_attr_refined$PRMS_segid == '3_1']
[1] 198.3
> p2_nhdv2_attr_refined$TOT_BASIN_AREA[p2_nhdv2_attr_refined$PRMS_segid == '3_2']
[1] 434.89
> p2_nhdv2_attr_refined$TOT_BASIN_AREA[p2_nhdv2_attr_refined$PRMS_segid == '8_1']
[1] 88.1
> p2_nhdv2_attr_refined$TOT_BASIN_AREA[p2_nhdv2_attr_refined$PRMS_segid == '8_2']
[1] 213.31
> p2_nhdv2_attr_refined$TOT_BASIN_AREA[p2_nhdv2_attr_refined$PRMS_segid == '51_1']
[1] 728.2
> p2_nhdv2_attr_refined$TOT_BASIN_AREA[p2_nhdv2_attr_refined$PRMS_segid == '51_2']
[1] 11760.08
```

And catchment areas are all small. I'd expect the TOT area for the _2 to be the TOT area for _1 + CAT area for _2 + a small error. Should we correct these TOT areas by computed them from the CAT areas?
```
> p2_nhdv2_attr_refined$CAT_BASIN_AREA_sum[p2_nhdv2_attr_refined$PRMS_segid == '3_2']
[1] 23.28
> p2_nhdv2_attr_refined$CAT_BASIN_AREA_sum[p2_nhdv2_attr_refined$PRMS_segid == '8_2']
[1] 2.92
> p2_nhdv2_attr_refined$CAT_BASIN_AREA_sum[p2_nhdv2_attr_refined$PRMS_segid == '51_2']
[1] 0.58
```

msleckman:
>  started by looking at the areas and was surprised that the TOT area is so much larger for _2 compared to _1 segments.

Does `_2` include `_1`

2_process/src/process_nhdv2_attr.R
```diff
@@ -269,3 +269,184 @@ process_catchment_nhdv2_attr <- function(file_path,vars_table,segs_w_comids,nhd_
...
+    STR_DENS_segs <- filter(nhdv2_attr_refined, hru_segment %in% ind_segs) %>% 
+      pull(CAT_STRM_DENS_area_wtd)
+    
+    #Compute TOT values
+    #need special handling for PRMS segments with _2
+    ind_replace <- which(nhdv2_attr_refined$hru_segment == subseg)
+    nhdv2_attr_refined$TOT_CWD_frmCAT[ind_replace] <- compute_tot(CWD_segs,areas_segs)
+    nhdv2_attr_refined$TOT_TAV7100_ANN_frmCAT[ind_replace] <- compute_tot(TAV_segs,areas_segs)
+    nhdv2_attr_refined$TOT_TMIN7100_frmCAT[ind_replace] <- compute_tot(TMN_segs,areas_segs)
+    nhdv2_attr_refined$TOT_STRM_DENS_frmCAT[ind_replace] <- compute_tot(STR_DENS_segs,areas_segs)
```
msleckman, line 392: Following my suggestion above, this can be built more concisely, if time allows, by looping through the new `TOT_frmCAT` cols.

2_process/src/process_nhdv2_attr.R
```diff
@@ -269,3 +269,184 @@ process_catchment_nhdv2_attr <- function(file_path,vars_table,segs_w_comids,nhd_
...
+                                                              TRUE ~ get(attr_i))
+                               ) %>%
+    pull(attr)
+  return(nhdv2_attr_refined)
+}
+
+compute_tot <- function(cat_var, areas_segs){
+  #' 
+  #' @description Function to compute TOT from CAT variables
+  #'
```
msleckman, line 445:
```suggestion

```

2_process/src/process_nhdv2_attr.R
```diff
@@ -269,3 +269,184 @@ process_catchment_nhdv2_attr <- function(file_path,vars_table,segs_w_comids,nhd_
...
+  #' @param cat_var vector of CAT variables for current and upstream segments
+  #' @param areas_segs area of each segment in the same order as cat_vec
+  #'
+  #' @value Returns the area-weighted sum (TOT) variable
```
msleckman, line 449: isn't this an area-weighted mean, with area_segs as the weight.

msleckman (commented):
Did a cursory review and have not yet run locally.

2_process/src/process_nhdv2_attr.R
jds485: yes, this is the same problem as #126

2_process/src/process_nhdv2_attr.R
jds485: yes, thanks for catching the typo!

2_process/src/process_nhdv2_attr.R
jds485: That's definitely a better method. Should be faster to loop over the columns. I'll wait to make that edit until after your review

2_process/src/process_nhdv2_attr.R
```diff
@@ -269,3 +269,184 @@ process_catchment_nhdv2_attr <- function(file_path,vars_table,segs_w_comids,nhd_
...
+  #Gather the sum of NHD reach lengths in m
+  ind_areas$length_m <- 0
+  for (i in 1:nrow(ind_areas)){
+    #all NHD reaches for this PRMS segment
+    nhd_reaches <- filter(prms_nhdv2_xwalk,
+                          PRMS_segid %in% ind_areas$PRMS_segid[i]) %>%
+      select(comid_cat) %>%
+      str_split(., pattern = ';', simplify = T)
+    
+    ind_areas$length_m[i] <- filter(nhdv2_reaches,
```
lekoenig, line 337: `ind_areas$length_m` indicates meters but the NHDPlus attributes for length are given in units of KM (LENGTHKM). I see the comment below about the magnitude of the stream density estimates, but I think that is just because the basin areas are _very_ small and not because of incorrect units on the length attributes:

Given your comment I quickly checked a couple things. 1) the length values using `sf::st_length` (shown below as check_length_m):
```
> ind_areas
# A tibble: 5 x 4
  PRMS_segid CAT_BASIN_AREA_sum check_length_m length_m
  <chr>                   <dbl>          <dbl>    <dbl>
1 282_1                 0.0009          0.0634    0.064
2 287_1                 0.00230         0.0480    0.048
3 878_1                 0.00365         0.0620    0.062
4 892_1                 0.00470         0.0471    0.047
5 894_1                 0.0018          0.0526    0.053
```

2) how the calculated values match up with the values from `CAT_STRM_DENS_area_wtd`
trdens

From these checks, I think the calculation/units are correct but these catchments are wonky.

2_process/src/process_nhdv2_attr.R
```diff
@@ -269,3 +269,184 @@ process_catchment_nhdv2_attr <- function(file_path,vars_table,segs_w_comids,nhd_
...
+      pull(CAT_TAV7100_ANN_area_wtd)
+    TMN_segs <- filter(nhdv2_attr_refined, hru_segment %in% ind_segs) %>% 
+      pull(CAT_TMIN7100_area_wtd)
+    STR_DENS_segs <- filter(nhdv2_attr_refined, hru_segment %in% ind_segs) %>% 
+      pull(CAT_STRM_DENS_area_wtd)
+    
+    #Compute TOT values
+    #need special handling for PRMS segments with _2
+    ind_replace <- which(nhdv2_attr_refined$hru_segment == subseg)
+    nhdv2_attr_refined$TOT_CWD_frmCAT[ind_replace] <- compute_tot(CWD_segs,areas_segs)
```
lekoenig, line 389: Your `compute_tot()` function does the trick, but {stats} included in base R also has a function called `weighted.mean` where you specify the weights using an argument `w`.

2_process/src/process_nhdv2_attr.R
```diff
@@ -269,3 +269,184 @@ process_catchment_nhdv2_attr <- function(file_path,vars_table,segs_w_comids,nhd_
...
+    pull(attr)
+  return(nhdv2_attr_refined)
+}
+
+compute_tot <- function(cat_var, areas_segs){
+  #' 
+  #' @description Function to compute TOT from CAT variables
+  #'
+  #' @param cat_var vector of CAT variables for current and upstream segments
+  #' @param areas_segs area of each segment in the same order as cat_vec
```
lekoenig, line 447:
```suggestion
  #' @param areas_segs area of each segment in the same order as cat_var
```

2_process/src/process_nhdv2_attr.R
```diff
@@ -269,3 +269,184 @@ process_catchment_nhdv2_attr <- function(file_path,vars_table,segs_w_comids,nhd_
...
+  
+  #Compute TOT variables from PRMS CAT variables
+  # CWD, TAV7100, TMIN7100, STRM_DENS
+  #Add hru segment identifier to match with the recursive function output ID
+  nhdv2_attr_refined$hru_segment <- apply(str_split(nhdv2_attr_refined$PRMS_segid, 
+                                                    pattern = '_', simplify = T), 
+                                          MARGIN = 1, FUN = first) %>%
+    as.numeric()
+  
+  nhdv2_attr_refined$TOT_CWD_frmCAT <- 0
```
lekoenig, line 362: Is it safer to initiate these values with NA rather than 0?

2_process/src/process_nhdv2_attr.R
lekoenig, line 354: Do you mind adding a comment that explains why it's necessary to compute TOT variables from CAT for these attributes? That might be helpful for interpreting this code at some later date.

lekoenig (approved):
Looks good, and the pipeline builds for me! 🎉  I've made some minor comments and suggestions, but I'm not requesting any changes.

I think it makes sense to bundle all of the feature processing steps within the `refine_features()` function as you've done here. I liked how the `refine_from_neighbors` code was ported out into a separate function. I think we could similarly consider sweeping the "calculate TOT from CAT" and "calculate drainage density" code chunks into their own small functions to make `refine_features` more readily interpretable. However, I view that as an enhancement and I approve the PR as is.

2_process/src/process_nhdv2_attr.R
```diff
@@ -269,3 +269,184 @@ process_catchment_nhdv2_attr <- function(file_path,vars_table,segs_w_comids,nhd_
...
+                                          TRUE ~ CAT_BASIN_AREA_sum)) %>%
+    #drop PRMS area column
+    select(-AREASQKM_PRMS)
+  
+  #RECHG
+  #Change recharge for NA segment to the average of its neighbors (from_segs and to_seg)
+  #index to change
+  nhdv2_attr_refined$CAT_RECHG_area_wtd <- refine_from_neighbors(nhdv2_attr_refined,
+                                                                 attr_i = 'CAT_RECHG_area_wtd', 
+                                                                 prms_reach_attr)
```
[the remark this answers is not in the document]
jds485: The necessary information exists in both dfs, so I changed to using p2_prms_reach_attr_df

2_process/src/process_nhdv2_attr.R
jds485: Yes, I also think that the length units are in km. The ScienceBase attribute description for stream density is the length in m / area in km^2, and it seems instead that it's length in km / area in km^2. I changed the variable name to `length_km` and edited the comment

jds485:
Thanks @msleckman  and @lekoenig! I think I've addressed all of your comments. I'm going to wait to merge this until after #126 is addressed because it may require adding some special case handling in the functions I wrote in this PR (looping over columns when computing TOT from CAT variables, refining from neighbors)

2_process/src/process_nhdv2_attr.R
```diff
@@ -269,3 +269,175 @@ process_catchment_nhdv2_attr <- function(file_path,vars_table,segs_w_comids,nhd_
...
+    STR_DENS_segs <- filter(nhdv2_attr_refined, hru_segment %in% ind_segs) %>% 
+      pull(CAT_STRM_DENS_area_wtd)
+    
+    #Compute TOT values
+    #need special handling for PRMS segments with _2
+    ind_replace <- which(nhdv2_attr_refined$hru_segment == subseg)
+    nhdv2_attr_refined$TOT_CWD_frmCAT[ind_replace] <- weighted.mean(CWD_segs,areas_segs)
+    nhdv2_attr_refined$TOT_TAV7100_ANN_frmCAT[ind_replace] <- weighted.mean(TAV_segs,areas_segs)
+    nhdv2_attr_refined$TOT_TMIN7100_frmCAT[ind_replace] <- weighted.mean(TMN_segs,areas_segs)
+    nhdv2_attr_refined$TOT_STRM_DENS_frmCAT[ind_replace] <- weighted.mean(STR_DENS_segs,areas_segs)
```
msleckman, line 391: Another enhancement suggestion to reduce redundant lines of code. Could you use `purrr::map2()` to loop through two vectors simultaneously (TOT_frmCAT_cols and Cat_variables) and populate the appropriate columns with the right weighted mean values?  

`TOT_frmCAT_cols <- c('TOT_CWD_frmCAT', 'TOT_TAV7100_ANN_frmCAT', 'TOT_TMIN7100_frmCAT', 'TOT_STRM_DENS_frmCAT')`
`Cat_variables <- c(CWD_segs, TAV_segs, TMN_segs, STR_DENS_segs)

Although map2 might not work as needed since we are populating specific columns. Considered providing a suggestion until realizing this.

msleckman (approved):
Successfully ran this (with most recent commits) and approving the merged into main. Left 1 comment but realized that I noted that in my earlier review.

Issue #41, opened 2022-01-03
Create single time series for each reach/gage


lekoenig:
Aggregate NWIS uv/dv data as well as discrete data to create model-ready time series of specific conductance that amount to 1 value/segment/day.

Steps include:

- [x] Create aggregated dataset that represents a time series for each segment (that aggregates multiple sites that may exist along that segment)
- [x] Create unaggregated dataset that represents a time series for each lat/lon location (maintaining multiple sites along a segment)
- [x] Output a summary table that indicates the variability among individual gages within a segment
- [ ] EDA: time series plot for each reach, colored by unique lat-lon site on that segment (will use these plots to assess intra-segment variation in SC)

PR #128, opened 2022-03-24
41 aggr time series

lekoenig:
Addresses #41.

The code changes here create a new target `p2_SC_observations` that represents the daily specific conductance mean/min/max aggregated at the scale of a PRMS segment, or 1 value/segment/day. 

The `aggregate_observations()` function has an argument for `aggr_method` so that we can also retain the individual site locations if of interest (`aggr_method = "site"` will aggregate sites that may have different site id's but share geographic coordinates per the lat/lon values given). There is also an argument for `prefer_nwis_sites` that defaults to FALSE but allows us to preferentially retain and use the data from NWIS sites for reach-days where multiple data sources are available (e.g., WQP data and continuous data from NWIS).

Here's a preview of the new target:
```
> tar_load(p2_SC_observations)
> head(p2_SC_observations)
# A tibble: 6 x 9
  subsegid Date       mean_value min_value max_value n_value sd_value cv_value site_ids
  <chr>    <date>          <dbl>     <dbl>     <dbl>   <int>    <dbl>    <dbl> <chr>   
1 1_1      1985-06-13         53        53        53       1       NA       NA 01413085
2 1_1      1985-08-21         64        64        64       1       NA       NA 01413085
3 1_1      1985-12-03         51        51        51       1       NA       NA 01413085
4 1_1      1986-03-14         44        44        44       1       NA       NA 01413085
5 1_1      1986-05-19         52        52        52       1       NA       NA 01413085
6 1_1      1987-03-27         41        40        42       1       NA       NA 01413085
> 
```

### DIFF
--- a/2_process/src/aggregate_observations.R
+++ b/2_process/src/aggregate_observations.R
@@ -1,0 +1,150 @@
+combine_wq_data <- function(wqp_data, nwis_data){
+  #' 
+  #' @description Function to combine WQP datasets with NWIS daily data
+  #' 
+  #' @param wqp_data data frame containing discrete water quality data 
+  #' from the wQP. Must contain cols MonitoringLocationIdentifier, 
+  #' OrganizationIdentifier, ActivityStartDate, resultVal2, 
+  #' ResultStatusIdentifier
+  #' @param nwis_data data frame containing combined instantaneous (uv)
+  #' and daily (dv) data; expected cols are agency_cd, site_no, Date, 
+  #' Parameter, Value, Value_cd, Value_Max, Value_Max_cd, Value_Min,
+  #' Value_Min_cd, dat_src
+  #' 
+  #' @value returns a data frame containing discrete and continuous 
+  #' water quality observations
+  #' 
+  
+  # Reformat WQP data to match NWIS datasets
+  wqp_data_formatted <- wqp_data %>%
+    # format site names for NWIS sites with prefix "USGS"
+    mutate(site_id = if_else(grepl('USGS',MonitoringLocationIdentifier),
+                             substr(MonitoringLocationIdentifier, 6, 100),
+                             MonitoringLocationIdentifier),
+           Parameter = "SpecCond") %>%
+    # harmonize column names with cols used in nwis_data
+    rename(agency_cd = OrganizationIdentifier,
+           site_no = site_id,
+           Date = ActivityStartDate) %>%
+    # find the min/max and the mean (if multiple observations exist 
+    # for a unique site-date)
+    group_by(site_no, Date, agency_cd, Parameter) %>%
+    summarize(Value = round(mean(resultVal2, na.rm = TRUE), 1),
+              Value_cd = unique(ResultStatusIdentifier),
+              Value_Max = max(resultVal2, na.rm = TRUE),
+              Value_Max_cd = NA_character_,
+              Value_Min = min(resultVal2, na.rm = TRUE),
+              Value_Min_cd = NA_character_,
+              .groups = "keep") %>%
+    ungroup() %>%
+    # pare down columns of interest and specify the data source
+    select(agency_cd, site_no, Date, Parameter, Value, Value_cd, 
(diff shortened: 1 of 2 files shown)


2_process/src/aggregate_observations.R
```diff
@@ -0,0 +1,150 @@
...
+  #' @param nwis_data data frame containing combined instantaneous (uv) and daily (dv) data; 
+  #' expected cols are agency_cd, site_no, Date, Parameter, Value, Value_cd, Value_Max, 
+  #' Value_Max_cd, Value_Min, Value_Min_cd, dat_src
+  #' @param sites_w_segs data frame containing site locations and matched segment id's;
+  #' must contain column site_id
+  #' @param aggr_method character vector indicating how observations should be aggregated;
+  #' options include "reach" and "site" (defaults to "reach"). If "reach", then values 
+  #' returned represent 1 value/segment/day, if "site", then values will be returned for
+  #' each lat/lon location (maintaining multiple sites along a segment) amounting to 
+  #' 1 value/site/day
```
jsadler2, line 73: Could you say why we need this? Something like this (is this true)??
```suggestion
  #' 1 value/site/day. Aggregating by "site" can be useful because occasionally there are multiple sites that have the same lat and lon. 
```

2_process/src/aggregate_observations.R
```diff
@@ -0,0 +1,150 @@
...
+                site_ids = paste0(unique(site_no), collapse = ","),
+                .groups = "keep") %>%
+      # suppress any warnings related to taking min/max summaries
+      # when no non-NA data exist for a reach-date; in these cases, 
+      # Inf will be returned for min_value, max_value
+      suppressWarnings() %>%
+      mutate(cv_value = (sd_value/mean_value)) %>%
+      ungroup()
+    
+  }
```
jsadler2, line 138: Could you do just split on the grouping but have the same `summarise` step those lines are the same? Warning: I haven't tried this and I'm a `dplyr` newbie.
```suggestion
    grouped_obs <- obs_data_w_segs %>%
      group_by(subsegid, Date) %>%
      # If prefer_nwis_sites = TRUE, check whether data for that segment comes from
      # multiple sources (i.e., WQP and NWIS). If multiple sources, filter out any
      # non-NWIS samples for that segment-date; otherwise, retain all samples
      {if(prefer_nwis_sites == "TRUE"){
        filter(.,if(n_distinct(dat_src) > 1) dat_src == "NWIS" else TRUE)
      } else {.}
  } else {
    
    # For aggr_method = "site", group by lat/lon since in the WQP database, there
    # may exist different site names that share the same geographic coordinates
    grouped_obs <- obs_data_w_segs %>%
      group_by(lat,lon,Date) 
  }
  
  obs_data_aggr <- grouped_obs %>%
     summarize(mean_value = round(mean(Value, na.rm = TRUE), 1),
                min_value = min(c(mean_value,Value_Min), na.rm = TRUE),
                max_value = max(c(mean_value,Value_Max), na.rm = TRUE),
                sd_value = round(sd(Value, na.rm = TRUE), 2),
                n_value = length(!is.na(Value)),
                site_ids = paste0(unique(site_no), collapse = ","),
                .groups = "keep") %>%
      # suppress any warnings related to taking min/max summaries
      # when no non-NA data exist for a reach-date; in these cases, 
      # Inf will be returned for min_value, max_value
      suppressWarnings() %>%
      mutate(cv_value = (sd_value/mean_value)) %>%
      ungroup()
```

jsadler2 (approved):
Looks good, @lekoenig. I have a couple of optional suggestions to consider.

2_process/src/aggregate_observations.R
lekoenig: Yeah, good idea! I added some text in the documentation here to explain that a user might just want to retain the unaggregated, raw data but also, we aggregate by 'site' because there are sometimes multiple site names for the same lat/lon.

lekoenig:
Thanks for your review, @jsadler2! I've accepted both of your suggestions in the revised code.

PR #135, opened 2022-04-04
Combine lulc data

msleckman:
Opening this PR to prep for merge after #134. 

Will need quick updating based on changes from #134 before merging. 

- [x] check how the 2000 (FORE-SCE) and 2001 (NLCD) data compare across all reaches


msleckman:
## Comparing 2000 FORESCE and 2001 NLCD:

Taking a sample of PRMS_segid ([gist here](https://gist.github.com/msleckman/fe437d59dda0f87b1cbf7ad576634e5a)) 
image

It looks like what has been classified as 2 in FORESCE is categorized as 2 in the NLCD legend classification. 

Propose changing from 10 classes to 9 and combining what falls into category 2 and category for NLCD legend and FORESCE legend.

msleckman:
This PR is ready for review and merge. 

Added R script `clean_lulc_data_for_merge.R` with a fun that includes couple cleaning/munging steps to rbind FORESCE & NLCD (slightly over-complex because originally build using previous versions of the input targets). 

New targets added in `2_process.R` : 

* `p2_all_lulc_data_cat`
* `p2_all_lulc_data_tot` --> note: Not final as its input df `p2_PRMS_FORESCE_proportions_tot` has not been updated following the addition of the dissolved catchment polygons. 

Notice there is no `p2_all_lulc_data_acc` target for lulc because only NLCD has those attributes (found target `p2_PRMS_NLCD_lc_proportions_acc`).

2_process.R
```diff
@@ -326,6 +327,30 @@ p2_targets_list <- list(
       )}
   ),
   
+  # Combine NLCD and FORESCE
+  ## Cat
+  tar_target(
+    p2_all_lulc_data_cat,
```
jds485, line 333: The column names for land cover classes should have CAT or TOT so they can be distinguished from each other when they're eventually combined into one table. To match the NHD format: `CAT_prop_lcClass_#` 

I'm not sure where the best place to make that edit would be. Maybe better to add when they are combined instead of having to change something we've already reviewed?

jds485 (approved):
Looks good! I made a commit to remove a column that a recent PR added to the TOT table. I just have a small comment to move a function outside of another function. After that, you can merge

msleckman:
@jds485  addressed to col rename (CAT and TOT) in this PR finally, and also split out the function as you noted. This created some redundancy in my approach because the `clean_lulc_data_for_merge()` wrapper function had parameters that were used in the  `clean_dflist_element().`

Ended up creating essentially duplicated param with the time I had. definitely interested in feedback on how to simplify. but in all this run and creates our expected output.

jds485:
Thanks! You can merge this. I'm not seeing a quick way to change the wrapper function implementation, and I think it's a fine way to do it

PR #146, opened 2022-04-13
NA handling for FORESCE

### LINKED ISSUE #138: Compute TOT FORESCE from CAT data
Use the update reach network table to update the computed TOT FORESCE land cover data

jds485:
This PR edits the handling of catchments with NA FORE-SCE data (due to being in the estuary or being on the edges of the DRB region). 

Looking at NLCD data, the estuary is classified as "water" instead of NA. So, to be consistent with NLCD, I've changed the NA data to "water" for all of the HRUs in the estuary. These HRUs were identified by an automatic filter and by manual inspection.

For all boundary NAs, areas are removed from the HRU, and the non-NA land cover proportions are recomputed as a proportion of the non-NA area. This assumes that the NA areas have the same proportional land cover as the non-NA areas.

3 HRUs with only NA values are dropped. The PRMS segments that matched to these HRUs thankfully have other HRUs, so the land cover proportion could still be computed for them. 

Closes #138


jds485:
Looks like I started from the synoptic sampling viz branch instead of from main. Let me close this PR and make a new PR that starts from main. Sorry about that!

PR #149, opened 2022-04-21
Fix pipeline warnings

### LINKED ISSUE #144: Warnings for across() and get_flowline_index()
1. Using `across()` in `filter()` is deprecated, use `if_any()` or `if_all()`
From @lekoenig: This is just an update to best practices wrt dplyr syntax. They introduced the if_any and if_all arguments so that you can filter based on whether all rows meet logical conditional statement versus any rows.

2. get_flowline_index(): search_radius units not set, trying units of points.
Can try fixing by setting the units using:
`search_radius = units::set_units(bird_dist_cutoff_m, "m")`

jds485:
This PR fixes 2 warnings:

1. search radius warning in `get_flowline_index`: Fixed by using `search_radius = units::set_units(search_radius*2, "m")` in the function call. I checked that it worked by loading in the old target as `test` and comparing the na locations and data values:
```
test <- tar_read(p2_sites_w_segs)
tar_make(p2_sites_w_segs)
tar_load(p2_sites_w_segs)
all(which(is.na(test), arr.ind = T) == which(is.na(p2_sites_w_segs), arr.ind = T))
[1] TRUE
all(test == p2_sites_w_segs, na.rm = T)
[1] TRUE
```
2. using across in filter: I changed to `if_all` for the one case we had of `across` in a `filter`. I tested that it worked using the same method as 1.  

Closes #144

### DIFF
--- a/2_process/src/match_sites_reaches.R
+++ b/2_process/src/match_sites_reaches.R
@@ -54,7 +54,7 @@ get_site_flowlines <- function(reach_sf, sites, sites_crs, max_matches = 1, sear
   flowline_indices <- nhdplusTools::get_flowline_index(flines = reaches_nhd_fields,
                                                        points = sites_sf,
                                                        max_matches = max_matches,
-                                                       search_radius = search_radius*2,
+                                                       search_radius = units::set_units(search_radius*2, "m"),
                                                        precision = 1) %>%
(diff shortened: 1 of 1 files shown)


jsadler2 (approved):
Looks good, @jds485

PR #156, opened 2022-05-18
Gridmet in pipeline

jsadler2:
Add gridmet file and associated targets to pipeline.


jds485:
Thanks! This looks good to me. Should this be merged after the S3 PR? It looks like it's built off of that branch

jsadler2:
Yep. We should merge the S3 PR first.

1_fetch.R
```diff
@@ -357,7 +378,22 @@ p1_targets_list <- list(
...
       },
-    format = "file"
+    format = "file",
+    repository = 'local'
+  ),
+
+  # CSV file of gridmet drivers aggregated to PRMS segments
+  tar_target(
+    p1_gridmet_csv,
+    "1_fetch/in/drb_climate_2022_04_06_segments.csv",
```
jds485, line 388: How large is this gridmet file? Would it be better to host it on S3?

jds485: Sorry, I meant to store the gridmet csv file on S3

jsadler2: I guess I'm not sure what you mean. Because it's a target, it _is_ stored on S3:

image

Is that what you mean? Or were you thinking of something else?

PR #161, opened 2022-06-09
160 s3 file fix

### LINKED ISSUE #160: Workaround to S3 file target tracking
S3 with file targets seems to work as expected to get files from S3 into the local repository.
But as [targets is designed for S3](https://github.com/ropensci/targets/discussions/851#discussioncomment-2913603), making a change to a file locally does not cause the file target to be outdated and the new data to be uploaded to S3. 

This is different behavior than if repository = local because then a change to the file causes the target to be outdated.

So, for now, I think we should make all files we track on GitHub have repository = local.

jds485:
Closes #160


jsadler2 (approved):
Looks good, @jds485.

I built this on a clean clone of the repo and it only built local targets that were not already existing in the repo.

jds485:
> I built this on a clean clone of the repo and it only built local targets that were not already existing in the repo.

Great! @lekoenig, @msleckman: If one of you has time to try building this, it's ready to be tested. You would test in a repo you've already built targets in with local storage. The switch to S3 storage should result in skipping all targets except the new local targets.

lekoenig:
I just tried running this branch on my local clone. One thing I'll note is that I did not have {paws} installed, so perhaps that package should be added in `tar_option_set`.

I wasn't exactly sure what to expect here, but when I run `tar_make()`, it does start building "non-local" targets like `p1_catchments_edited` and `p1_inst_data` that I was expecting to just pull down from S3. When I tried to visualize the network, most targets were outdated:
ntw_map

Is that what you expected, @jds485? This network map makes me wonder if {targets} has picked up on differences between my local clone and S3, but rather than _pulling_ those outdated targets, it appears to be re-building/writing them. Here's a snippet from my console:

```
> Sys.setenv(
+     AWS_PROFILE = "dev",
+     AWS_REGION = 'us-east-1',
+     AWS_SHARED_CREDENTIALS_FILE = "./credentials"
+ )
> tar_make()
* start target p1_dummy
* built target p1_dummy
v skip target p1_catchments_edited_gpkg
* start target p1_catchments_edited_sf
* upload _targets/scratch\p1_catchments_edited_sf2ab05ab21b75 part 1 of 2
* upload _targets/scratch\p1_catchments_edited_sf2ab05ab21b75 part 2 of 2
* built target p1_catchments_edited_sf
v skip target p1_nhdv2reaches_sf
v skip target p1_vars_of_interest_csv
v skip target p1_vars_of_interest
v skip target p1_natural_baseflow_zip
v skip target p1_natural_baseflow_csv
v skip target p1_NLCD2011_data_zipped
v skip target p1_FORESCE_backcasted_LC
v skip target p1_nwis_sites
* start target p1_nwis_sites_inst
* built target p1_nwis_sites_inst
* start branch p1_inst_data_e5909bc0
Retrieving instantaneous data for 01413500
* built branch p1_inst_data_e5909bc0
* start branch p1_inst_data_cfe667df

```

jds485:
I'm not sure why these targets would have been rebuilt. I just tried running my local copy of Margaux's fork and Lauren's fork of USGS-R main and it seems to work as I would have expected. The initial `tar_visnetwork()` showed a lot of outdated targets, similar to what you show in your comment. After running `tar_make()` (up to the target that I know will not work because it's fixed in this PR), all targets except the `repository = local` targets were skipped, and the `tar_visnetwork` showed targets as up to date even though they were simply skipped. 

The `_targets/objects` folder DID NOT update after this pipeline run, so it has outdated targets in it. That is something to be careful with if we switch a target from local to aws storage. The local target will never be accessed by the pipeline, but we can still access it so it's probably best to manually remove the local target.

It seems like file targets stored on S3 did not download when they were skipped.

jds485:
I clarified with Lauren that she ran the 160-S3-file-fix branch. I am not sure why the targets built for that branch. After checking with Lauren, I am merging this PR because it worked as expected for the other branches

PR #162, opened 2022-06-10
editing scientific notation options

### LINKED ISSUE #155: Make site numbers characters in log files
Sites with long numbers are saving in scientific notation. Example:
4.03447E+14
in `inst_timeseries_summary.csv`

jds485:
Edits options to ensure that the site numbers do not write in scientific notation. Decided this was a simpler fix than changing the number to a character.

I did not run this yet because these targets are changed to be stored locally in #161.

Closes #155


jds485:
Proof that this worked to retain all digits within the scientific notation:
image

PR #163, opened 2022-06-10
download zip and unzip in one step

### LINKED ISSUE #61: p1_NLCD_data_zipped gets outdated after first run through pipeline
After a 1st `tar_make()`, three NLCD targets are outdated (image below). When a 2nd `tar_make()` is called, the first two (zipped and unzipped) are rebuilt, and the last is skipped. After that, `tar_make()` skips all targets. So, I'm expecting that something in the `p1_NLCD_data` function is causing the previous two NLCD targets to become outdated. I did not see anything that would cause it, though.
image

jds485:
This PR:
1. merges targets to download a zipped file and unzip that file in the same target. Applies to NLCD 2011 data and baseflow data.
2. edits functions to remove zipped files in the unzip step (saves storage space)

I have done a local check that this works, but I have not run the pipeline because some targets switch from aws to local in #161 

Closes #61


jds485:
Confirming that the pipeline works with these edits, and now does not need to rebuild these targets after they have been built once.  I'm going to self-merge this.

PR #165, opened 2022-06-10
add new date variables

### LINKED ISSUE #79: Edits to using dummy_date
`dummy_date` functions as the date we’re cutting off timeseries, and also forces rebuilds of targets. But we didn’t consider before that a fresh pipeline rebuild with the same dummy_date would generate targets on a different date than the dummy_date.

Suggestion is to add a new variable for the end date of timeseries, and another variable for the date the pipeline is created.

jds485:
This PR adds a `latest_date` for which we want NWIS data, and retains the `dummy_date` as a way to update the NWIS data downloads. I set the date of `dummy_date` to the date of last download on S3. I think the use of `dummy_date` is deprecated because we're now using `tar_cue` to never download NWIS data (until we want to download it again). But I'm keeping it in the pipeline in case we move away from S3 in the future.

I have not run this yet. Waiting on other open PRs

Closes #79


jds485:
Changing the date variables caused the NWIS download targets to be outdated. We previously used tar_cue to never re-download NWIS sites, so I decided to add a cue = never to never re-download the daily and instantaneous data (e.g., when a variable is changed). This way, we're in control of when these data are updated.

1_fetch.R
```diff
@@ -11,6 +11,9 @@ source("1_fetch/src/fetch_nhdv2_attributes_from_sb.R")
 source("1_fetch/src/download_file.R")
 source("1_fetch/src/munge_reach_attr_tbl.R")
 
+# tar_cue for downloading NWIS sites and data
```
lekoenig, line 14: I understand wanting peace of mind that the NWIS targets haven't been rebuilt without our realizing it. If we _did_ decide to re-build those targets, what would we change `NWIS_cue` to? `always`? A small note in the comment here might be useful for indicating that.

_targets.R
```diff
@@ -89,11 +89,12 @@ GFv1_HRUs_edited_url <- "https://github.com/USGS-R/drb-network-prep/blob/940073e
...
 stat_cd_select <- c("00001","00003")
 
-# Define earliest startDate for NWIS data retrievals
+# Define earliest startDate and latest endDate for NWIS data retrievals
 earliest_date <- "1979-10-01"
+latest_date <- "2021-12-31"
 
 # Change dummy date to force re-build of NWIS SC sites and data download
-dummy_date <- "2021-12-31"
+dummy_date <- "2022-06-16"
```
lekoenig, line 97: Given the addition of `NWIS_cue` in this PR, do you expect that `dummy_date` ever gets used in the pipeline? Should this variable be omitted, or is it still useful as a sort of documentation?

lekoenig (approved):
I haven't run these code changes, but I added a couple comments to this PR.

1_fetch.R
jds485: Thanks, I'll add that comment. It would change to 'thorough'

_targets.R
jds485: If we remember to change this whenever we change the NWIS_cue, then it does serve as documentation of the date we last updated the NWIS data. I'm going to move the dummy_date to the line below NWIS_cue so that it encourages us to remember.

lekoenig: Good idea, and is there a reason `NWIS_cue` is defined in 1_fetch.R rather than with the other variables in _targets.R?

jds485: The pipeline did not build when I had NWIS_cue in _targets.R. I'm not sure why that resulted in a variable not found error. It works when it's in 1_fetch.R

PR #167, opened 2022-06-10
remove NA filter for stream density

### LINKED ISSUE #166: Compute stream density for all PRMS catchments
We currently compute stream density for PRMS catchments with NA values. We will be updating the flow network to remove divergences, so the stream density for all catchments should be updated to match the new network file.

jds485:
This PR computes CAT stream density for all PRMS segments based on the NHD flowlines in the PRMS catchment (comid_cat). We previously computed these values for the catchments with NA stream density, so the edit removes that NA filter.

I have not run this yet. Waiting on other open PRs.

Closes #166


jds485:
For most catchments, the change in stream density is less than 5% of the new value. The one site with a 20% change corresponds to the smallest catchment area, so even a small change in the length of streams will cause a large change in the stream density.
image

PR #173, opened 2022-06-28
145 update network

### LINKED ISSUE #145: Add handling for NHD reaches with AREASQKM==0
Two suggestions:
1. We currently compute an approximate area for these reaches using the length^2, but we could just keep using 0 area so that the data for these 0 area reaches are not counted toward to final area-weighted PRMS attribute value. However, some PRMS reaches will then not receive attributes because the whole area is 0, so some special handling could be needed for these reaches. (PRMS segs “287_1”, “614_1”, “878_1”,”903_1”). Also may not work for attributes that do not use an area-weighted value.
2. filter 0 area reaches out of `p1_nhdv2reaches_sf` and check how that impacts downstream targets. Some PRMS segments may not have any NHD reaches after this filter.

jds485:
This PR updates the NHD to PRMS crosswalk by using an NHD network that omits zero area reaches.

Closes #145


lekoenig (approved):
I have not re-run the full pipeline, but the commit hashes reflect the updates made in `drb-network-prep` to add a version of the crosswalk table with divergences but without flowline reaches where `AREASQKM == 0`. 

With these changes, subsegid "287_1" should be the only segment with a COMID where `AREASQKM == 0` because that COMID is the only one that makes up subseg 287_1:
```r
> tar_load(p2_drb_comids_all_tribs)
> tar_load(p1_nhdv2reaches_sf)
> 
> all_tribs_flines <- p2_drb_comids_all_tribs %>%
+     left_join(p1_nhdv2reaches_sf %>%
+                   select(COMID, AREASQKM) %>%
+                   mutate(COMID = as.character(COMID)) %>%
+                   sf::st_drop_geometry(),
+               by = c("comid"="COMID")) %>%
+     filter(AREASQKM == 0)
> 
> all_tribs_flines
# A tibble: 1 x 3
  PRMS_segid comid   AREASQKM
  <chr>      <chr>      <dbl>
1 287_1      4188275        0
>
```

lekoenig:
> Thanks! Did you load p2_drb_comids_all_tribs and p1_nhdv2reaches_sf from my 145-update-network branch?

I pulled down the PR and ran those targets using repository = 'local' to generate the code snippet above.

jds485:
Okay, thanks! I think you could have used the aws repository. The targets should have pulled from there. Or maybe this is back to the issue you had of switching from local to aws on this branch. Anyway, thanks for the review!

PR #175, opened 2022-07-08
Adds dynamic from static attributes and lagged dynamic attributes

### LINKED ISSUE #117: Create dynamic features from static attributes
Several static features (CAT and TOT) can be converted into dynamic features that are assigned to observations by date:
- Housing density - decadal
- Number of major dams  - decadal 
- Total number of dams (Ndams)  - decadal
- Dam NID storage  - decadal
- Dam NORM storage  - decadal
- Land cover - current and previous in 3 year increments back 21 years
- gridMET - current day and previous day, week, month, 3 months, year. We lose SC data the further back we go. Example: for 1 year, we’d start modeling in 1980 instead of 1979. 

The date of record can be treated as the center value.  Example for decadal data: an observation on 12/31/1985 could be assigned the 1980 value, and on 1/1/1986 assigned the 1990 value. 

Ends of time series can be assigned the most recent value

I think a generic function could be created to handle assigning these attributes to observations based on the date.

The land cover portion of this is related to #102. The result of #102 would be the current land cover timeseries.

After the timeseries are created, the CAT and TOT variables for housing density and dams can be removed from `p2_nhdv2_attr_refined`. (land cover is in a separate target)

jds485:
This PR computes dynamic attributes from static attributes listed in #117. It also computes lagged values using either the mean over the specified lag period or the exact value on the lagged day. There is a target that allows for specification of lags to compute using the lubridate formats.

The dynamic data are computed for every day from the start_date to the end_date to allow for predictions on site-days without SC observations. Static data are joined to these data by PRMS segid. A target is also available for only the SC site-days, which will be used in model training. Note that some rows have NA for lagged information attributes (gridmet and baseflow lags up to 5 years).

The largest target containing all dates and all attributes is about 7.5 GB. It will take several minutes to load in. I'd recommend instead inspecting `p2_all_attr_SC_obs` which is < 250 MB.

Additional edits:
The `p2_rdsalt_per_catchment_allyrs` target was added to the `p2_nhdv2_attrs` target and named CAT_rdsalt_prop. A TOT attribute is now computed for road salt in the `process_nhdv2_attr` function.

One NA reach for baseflow data was filled in from the average value for neighboring reaches. Averaging was weighted by drainage area.

Note:
It takes about 3 hours to run, mostly from the lag computation. I'd be interested in your thoughts on speeding that up. It might be possible to add in a parallelization over the lags using foreach %dopar%, for example

Closes #117


2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+    filter(attribute == lag_attr_name) %>%
+    pull(lag_fxns) %>%
+    unlist()
+  
+  if (length(tmp_lag_fxns) > 0){
+    #Loop over all functions except exact
+    for (f in 1:length(tmp_lag_fxns)){
+      if (tmp_lag_fxns[f] != 'exact'){
+        #Loop over all lags
+        for (lag in 1:length(tmp_lags)){
```
jds485, line 741: Also could use foreach here

2_process/src/process_nhdv2_attr.R
```diff
@@ -552,3 +557,82 @@ refine_from_neighbors <- function(nhdv2_attr, attr_i, prms_reach_attr
...
+                           seg_match == '51_1' ~ '51_1',
+                           seg_match == '51_2' ~ '51_2',
+                           TRUE ~ paste0(seg_match, '_1'))
+    
+    #get the average of the attributes for the matched reaches
+    reach_vals <- filter(dyn_attr, PRMS_segid %in% seg_match)
+    #check if any are equal to exactly 0
+    if (any(reach_vals == 0, na.rm = TRUE)){
+      warning('some neighboring reaches of reach ', ind_reach[j], 
+              'have a value of 0 for ', attr_i)
```
lekoenig, line 606: I like these checks! It's helpful to know what the values look like when we impute missing observations using the neighboring reaches.

2_process/src/process_nhdv2_attr.R
```diff
@@ -552,3 +557,82 @@ refine_from_neighbors <- function(nhdv2_attr, attr_i, prms_reach_attr
...
+    }
+    #check if all reaches have NA values
+    if (all(is.na(reach_vals))){
+      warning('all neighboring reaches of reach ', ind_reach[j], 
+              'have a value of NA for ', attr_i, '. This reach will still be NA.')
+    }
+    
+    #need to group by date and compute the mean/total drainage area
```
lekoenig, line 614: Does this comment refer to the mean baseflow column or the mean drainage area among neighboring reaches?

2_process.R
```diff
@@ -448,7 +454,11 @@ p2_targets_list <- list(
   # Create combined NHDv2 attribute data frame that includes both the cumulative upstream and catchment-scale values
   tar_target(
     p2_nhdv2_attr,
-    create_nhdv2_attr_table(p2_nhdv2_attr_upstream, p2_nhdv2_attr_catchment)
+    create_nhdv2_attr_table(p2_nhdv2_attr_upstream, p2_nhdv2_attr_catchment) %>%
+      #add CAT road salt to static attributes
+      left_join(p2_rdsalt_per_catchment_allyrs %>% 
+                  select(PRMS_segid, rd_salt_all_years_prop_drb) %>%
+                  rename(CAT_rdsalt_prop = rd_salt_all_years_prop_drb), by = 'PRMS_segid')
```
lekoenig, line 461: Here, the road salt values are appended to the data frame that contains all of the attributes summarized from the datasets on ScienceBase. Are there other datasets that were calculated by raster processing that could also be joined here? Or is land cover the only one (and land cover you are only treating as a dynamic attribute, rather than as a static attribute)?

lekoenig (commented):
This PR is pretty big so I will review it in chunks: neighboring reaches imputation, road salt steps, calculating lagged values, static --> dynamic

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+  
+  #make a dynamic features dataframe
+  dates <- seq(as.Date(start_date), as.Date(end_date), 1)
+  segs <- sort(unique(attrs$PRMS_segid))
+  df <- tibble(Date = dates, seg = segs[1])
+  for(i in 2:length(segs)){
+    df_tmp <- tibble(Date = dates, seg = segs[i])
+    df <- rbind(df, df_tmp)
+  }
+  rm(df_tmp, dates, i)
```
lekoenig, line 35: This code chunk probably isn't the lowest-hanging fruit for optimization, but storing `df_tmp` and iteratively appending it to `df` may be somewhat costly. I suggest one way of speeding this up below:

```r
> start_time <- Sys.time()
> 
> df <- tibble(Date = dates, seg = segs[1])
> for(i in 2:length(segs)){
+     df_tmp <- tibble(Date = dates, seg = segs[i])
+     df <- rbind(df, df_tmp)
+ }
> end_time <- Sys.time()
> elapsed <- end_time - start_time
> print(elapsed)
Time difference of 47.50728 secs
> 
> 
> start_time <- Sys.time()
> 
> df_new <- do.call(rbind, lapply(1:length(segs), function(i) 
+     tibble(Date = dates, seg = segs[i])))
> 
> end_time <- Sys.time()
> elapsed_new <- end_time - start_time
> print(elapsed_new)
Time difference of 6.094147 secs
>
> identical(df, df_new)
[1] TRUE
> 
```

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+                                                      start_date = start_date) %>%
+                  select(-Date, -seg, -all_of(col_select))
+    )
+  }
+
+  return(df)
+}
+
+get_four_digit_year <- function(two_digit_yr){
+  #' @description computes the 4 digit year from two digit year
```
lekoenig, line 448: It might be useful to add to this function description that `get_four_digit_year()` is specific to the `HDENS` data because of the range in year included in that dataset. For example, if an attribute dataset included data from 1930, this function would not work as intended.

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+  #Special handling for each variable
+  if ('HDENS' %in% dyn_cols){
+    #Housing density has the format HDENSYY
+    tmp_attrs <- select(attrs, PRMS_segid, contains('HDENS'))
+    
+    #Convert the column names to 4-digit years
+    col_years <- str_split(colnames(tmp_attrs)[-1], 
+                           pattern = '_', simplify = TRUE)[,2] %>%
+      substr(start = 6, stop = 7) %>%
+      get_four_digit_year()
```
lekoenig, line 46: This works, but it assumes that the structure of the HDENS names will always be the same. I don't think we'd expect any changes, but nonetheless it might be more robust to replace `col_years` with:

```r
   col_years <- stringr::str_extract(names(tmp_attrs)[-1], "[[:digit:]]+") %>%
      get_four_digit_year()
```

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+    
+    #HDENS data date ranges and dataset years
+    tmp_date_ranges <- seq(as.Date('1965-09-30'), as.Date('2005-09-30'),
+                           '10 years')
+    tmp_years <- tmp_attrs %>% 
+      select(starts_with(attr_prefix[1])) %>% 
+      colnames() %>% 
+      sort() %>% 
+      str_split(pattern = '_', simplify = TRUE) %>% 
+      .[,2]
```
lekoenig, line 65: Similar to my comment above, I think extracting the digits rather than specific indices within the `str_split` output is a little easier to read and interpret.

```r
tmp_years <- tmp_attrs %>% 
      select(starts_with(attr_prefix[1])) %>% 
      colnames() %>% 
      stringr::str_extract("[[:digit:]]+") %>%
      sort()
```

2_process/src/process_nhdv2_attr.R
```diff
@@ -552,3 +557,82 @@ refine_from_neighbors <- function(nhdv2_attr, attr_i, prms_reach_attr
...
+  ind_reach <- filter(dyn_attr, is.na(get(attr_i))) %>%
+    pull(PRMS_segid) %>%
+    unique()
+  
+  for (j in 1:length(ind_reach)){
+    #find the from and to segments for this reach
+    seg_match <- filter(prms_reach_attr, PRMS_segid_main == ind_reach[j]) %>%
+      select(from_segs, to_seg) %>%
+      unique() %>%
+      mutate(segs = list(c(from_segs[[1]], to_seg))) %>%
```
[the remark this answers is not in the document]
jds485: `from_segs` is a one element list, and that list has a vector of seg IDs. So yes, using [[1]] will apply as long as the format we're using for `from_segs` does not change. `unlist()` also works

2_process.R
jds485: I debated also joining the 2011 riparian buffer attributes, but that dataset is currently not processed to PRMS segments so I decided to omit it for now. The other land cover data do not need to be merged here because they're handled dynamically. So I don't think there are others that would need to be added here, but thanks for checking!

2_process/src/process_nhdv2_attr.R
jds485: I should move this comment within the `drainage_area` if statement. When `drainage_area` is provided, then the values are normalized by drainage area. I used that for baseflow because flow scales well by drainage area and it made more sense than taking a simple average.

2_process/src/add_dynamic_attr.R
jds485: Thanks! I'll make this edit

2_process/src/add_dynamic_attr.R
jds485: I'll edit this for all of the attributes

lekoenig (commented):
There are a couple ways we could make building `p2_dyn_attr` more efficient. One is to optimize the code within the functions. Beyond what you've proposed, I don't really have great ideas for ready-to-go implementations that would speed up the steps in `compute_lagged_attrs()`.

Another way the code could be made more efficient is to use `targets` functionality to reduce unnecessary rebuilds of this target which takes a long time to build. For example, if I remove 'HDENS' from the list of `dyn_cols` in `p2_dyn_attr`, that would trigger a rebuild of the _entire_ target, even though we've already built the others before! Similarly, if I added one lag value anywhere in `p2_dyn_attr_lags` we'd have to start fresh for _all_ of the `dyn_cols`. 

I have not tried myself to implement branching for `p2_dyn_attr`, but, for example, could you create a vector of `dyn_cols` in `_targets.R` and then branch over those cols while building `p2_dyn_attrs`? I could envision adding similar functionality for lags and then using `targets::cross()` to build specific slices of dyn-cols x lag values only when updates to those slices are needed. Maybe the formatting of `p2_dyn_attr_lags` makes that latter option a little tougher.

Branching is conceptually straightforward when we are combining data frames by binding rows, but in our pipeline you can see `add_dyn_attrs_to_reaches` and `p2_nhdv2_attr` for examples of how we've combined columns across branches before.

I think it's also OK to tackle these optimization steps in a separate "enhancement" issue that can be addressed as time allows.

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+    #Create a temporary date column containing the lagged date to use
+    if(lags[lag] != 0){
+      dyn_df$lagged_Date <- dyn_df$Date %m-% period(paste(lags[lag], lag_unit[lag]))
+    }else{
+      dyn_df$lagged_Date <- dyn_df$Date
+    }
+    
+    #Join data to all segs by date range
+    # Note: using this method instead of a case_when because
+    # case_when was very slow
```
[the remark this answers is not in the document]
jds485: Thanks! I agree that the case_when would be easier to read. I programmed it that way first and it took a lot longer than this method. I like your suggestion to make a separate issue for optimizing this code. I can grab your suggestions into an issue on that

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+            left_join(dyn_df %>% 
+                        filter(lagged_Date <= date_ranges[i]) %>% 
+                        select(seg), 
+                      attrs %>% 
+                        select(PRMS_segid, 
+                               paste0(attr_prefix[j], '_', attr_years[i])),
+                      by = c('seg' = 'PRMS_segid')) %>%
+            select(paste0(attr_prefix[j], '_', attr_years[i]))
+        }else if (i == (length(date_ranges)+1)){
+          dyn_df[dyn_df$lagged_Date > date_ranges[i-1],
```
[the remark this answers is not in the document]
jds485: I'll add a comment for this. It's there because the `date_ranges` specify transition dates between statis attribute datasets. So the `date_ranges` vector is 1 element shorter than the total number of static attribute datasets that are made into a dynamic timeseries

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+  #' @description computes dynamic attribute from the provided tbl of static
+  #' attributes.
+  #' 
+  #' @param dyn_df tbl containing "Date" and the PRMS "seg". Should have all
+  #' dates that you want to use in modeling. 
+  #' @param attrs tbl containing the "PRMS_segid" and attribute columns
+  #' in the format <attr_prefix>_YYYY
+  #' @param lag_table table with 3 columns: attributes, lags, and lag_unit.
+  #' This is used to compute the lagged attributes. Can set to NULL to use a 
+  #' lag of 0.
```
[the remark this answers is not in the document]
jds485: I can add to this description. When `lag_table` is specified, then attributes are computed for each of the lags. Within that table, one of the `lags` could be 0

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+                                               filter = rep(1/as.numeric(days_lag), days_lag), 
+                                               sides = 1, 
+                                               .names = paste0("{.col}_", tmp_lags[lag], 
+                                                               tmp_lag_unit[lag], "_mean")), 
+                                        .groups = 'drop') %>%
+                              select(-seg) %>%
+                              #need to shift the results by 1 day because filter 
+                              #includes the current day in the lag
+                              add_row(.before = 1) %>%
+                              slice(-nrow(.))
```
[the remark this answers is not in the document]
jds485: Thanks! I appreciate all of the optimization suggestions. I might try the foreach loop because I'll have to rebuild these targets after making the PR changes, anyway. And I know that those loops are the main time consuming steps

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+  }
+  tmp_lag_fxns <- lag_table %>% 
+    filter(attribute == lag_attr_name) %>%
+    pull(lag_fxns) %>%
+    unlist()
+  
+  if (length(tmp_lag_fxns) > 0){
+    #Loop over all functions except exact
+    for (f in 1:length(tmp_lag_fxns)){
+      if (tmp_lag_fxns[f] != 'exact'){
```
lekoenig, line 739: Same comment as above regarding the use of `exact` in a nested if statement.

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+  
+  #Remove the dates before the start_date
+  dyn_df <- dyn_df %>% 
+    filter(Date >= start_date)
+  
+  return(dyn_df)
+}
+
+
+compute_lagged_attrs_from_dynamic <- function(dyn_df, lag_table, lag_attr_name,
```
lekoenig, line 704: `compute_lagged_attrs()` and `compute_lagged_attrs_from_dynamic()` seem very similar. Are there certain steps that make it difficult to combine these into a single function to compute lagged attributes?

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+    pull(lag_unit) %>%
+    unlist()
+  #Make the length of the lag_unit match the length of lags
+  if ((length(tmp_lag_unit) == 1) & (length(tmp_lags) > 1)){
+    tmp_lag_unit <- rep(tmp_lag_unit, length(tmp_lags))
+  }
+  tmp_lag_fxns <- lag_table %>% 
+    filter(attribute == lag_attr_name) %>%
+    pull(lag_fxns) %>%
+    unlist()
```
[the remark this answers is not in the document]
jds485: That's a good catch. I initially had the `exact` computation as an `else` in this, and then decided to move it because it was faster to compute outside of this function.

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+    #Keeping this outside of the compute lagged attrs function because it's
+    #faster to compute this way.
+    if ('exact' %in% lag_table[lag_table$attribute == 'HDENS',]$lag_fxns[[1]]){
+      df <- get_dynamic_from_static(dyn_df = df, attrs = tmp_attrs, 
+                                        lag_table = lag_table,
+                                        date_ranges = tmp_date_ranges,
+                                        attr_years = tmp_years,
+                                        attr_name = 'HDENS', 
+                                        attr_prefix = attr_prefix)
+    }
```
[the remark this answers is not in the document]
jds485: This step computes the exact value on the day of the lag. It ended up being faster to do the `exact` computation outside of the `compute_lagged_attrs` function

lekoenig (commented):
I've now looked through all of the code in `add_dynamic_attr.R`. I focused more closely on 'HDENS' and 'baseflow' as examples of static and dynamic attributes for which we're calculating lagged values. 

I haven't inspected the outputs themselves very closely and am currently having odd issues connecting to S3 via saml2aws. Let me know if you'd like me to take a look at those and I can try again later.

2_process/src/add_dynamic_attr.R
```diff
@@ -0,0 +1,796 @@
...
+    if (!is.null(Upstream_Land)){
+      tmp_attrs_up <- pivot_wider(Upstream_Land %>% 
+                                    select(-PRMS_area_km2), 
+                                  names_from = Year, 
+                                  values_from = starts_with(Upstream_Land_prefix))
+    }
+    
+    #Loop over all land cover classes to assign attributes
+    for (lc in 1:9){
```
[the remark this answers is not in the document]
jds485: Those are the reclassified values. I can make this more generic by finding the land cover class numbers.

2_process/src/add_dynamic_attr.R
jds485: I'm not sure why I made these separate functions. There are only 2 differences! attr_name can be specified as '' to work for the from_dynamic version, and then the code at the end ` #Remove the dates before the start_date` can be included as an if statement. I'll change to make this one function. Thanks!

jds485:
I think I addressed your comments, @lekoenig. I have not rebuilt the targets yet

jds485:
The pipeline built overnight and produced the same attribute table after these code edits. The upload time is also a significant part of the runtime for these targets. Target build time was 2.45 hrs and the total time including upload to S3 was 3.3 hrs

2_process/src/add_dynamic_attr.R
lekoenig: That's great, thanks! I assumed I was just missing a more substantive difference between the two.

lekoenig (approved):
Thanks for these edits, @jds485! This PR looks good to me. I was able to build the pipeline with these changes using `repository = 'local'` and `NWIS_cue = 'thorough'`. Here is a preview of the output I see:

```r
> tar_load(p2_all_attr)
> dim(p2_all_attr)
[1] 7083747     504
> tar_load(p2_dyn_attr)
> dim(p2_dyn_attr)
[1] 7083747     246
>
```

That's a lot of attributes! 😃

PR #197, opened 2022-11-04
Add zarr file targets

### LINKED ISSUE #196: Add zarr file targets
Input data and observation file format needed for use in river-dl

jds485:
This PR adds 2 targets that generate zarr files of the attributes and SC observations, respectively. They are stored locally because targets did not allow saving to S3 as a directory (which zarr files are). We could probably edit to track each individual file, but that seems unnecessary given how we plan to use these files will require local storage, anyway.

Also includes an edit to the launch-rstudio....slurm script that was needed after a recent change to how rocker containers specify their paths. This was necessary to launch RStudio server.

There is also a new container for the pipeline that is needed to run these targets. It has Python and reticulate installed.

Closes #196


lekoenig (approved):
> There is also a new container for the pipeline that is needed to run these targets. It has Python and reticulate installed.

This looks good, @jds485! Did you have to add Python to the container to build these zarr targets, or was that for something else related to running river-dl?

PR #199, opened 2022-11-07
Add river adjacency matrix

### LINKED ISSUE #169: Update adjacency matrix
The matrix we pulled from a ScienceBase entry has 3 reaches missing. This issue is to generate a new adjacency matrix that contains those 3 reaches.

lekoenig:
This PR adds the subseg adjacency matrix from the DRB temperature quest to our project pipeline. `1_fetch/in/subseg_distance_matrix.rds` was downloaded from the shared temperature data cache and is committed here. The resulting object is a list containing four different versions of the distance matrix. I've created a separate target called `p1_drb_distance_matrix_updown` in case we opt to use other versions of the distance matrix down the line.

I haven't yet run the pipeline to push these targets to S3. @jds485, would you mind updating the targets meta file? Or should I go ahead and commit/push my version?

Closes #169


jds485:
Looks good! I updated the targets and also made a csv file of the distance matrix that can be read into Python. I think this can be merged after you review what I added

1_fetch.R
```diff
@@ -430,6 +430,37 @@ p1_targets_list <- list(
...
+  ),
+  #save as csv for use in Python
+  tar_target(
+    p1_drb_distance_matrix_updown_csv,
+    {
+      write.csv(x = as.data.frame(p1_drb_distance_matrix$updown), 
+              file = "1_fetch/out/drb_distance_matrix_updown.csv",
+              row.names = TRUE)
+      "1_fetch/out/drb_distance_matrix_updown.csv"
+    },
```
lekoenig, line 461: Oh, great addition - I wasn't thinking about the need to export for python!

One small comment that writing the file path twice increases the likelihood of making a typo - this looks good as you've written it but here's a proposed edit I think we've used elsewhere in this pipeline:

```suggestion
    {
      fileout <- "1_fetch/out/drb_distance_matrix_updown.csv"
      write.csv(x = as.data.frame(p1_drb_distance_matrix$updown), 
              file = fileout,
              row.names = TRUE)
      fileout
    },
```

jds485:
Thanks! Before merging, I want to double check the format expected by river-dl (I think it might be a numpy array instead of csv, and I want to also check the format of the rows and columns). I may add one more commit with a target that writes a numpy array.

jds485:
I added the npz file generation target

2_process/src/write_data.R
```diff
@@ -30,3 +30,34 @@ write_df_to_zarr <- function(df, index_cols, out_zarr) {
...
 }
+
+
+write_dist_matrix_npz <- function(mat_list, out_npz) {
+  #' @title Write an R list of matrices to a numpy npz store
+  #'
+  #' @description This function uses {reticulate} to write an R list of matrices to a numpy data
+  #' store (the file format river-dl currently takes). 
+  #'
+  #' @param mat_list list of matrices. Columns and rows must all be the same order in each matrix
```
lekoenig, line 41: What do you think of adding a brief comment here that we're expecting a list containing four matrices named "complete," "downstream," "upstream," and "updown?"

1_fetch.R
```diff
@@ -450,7 +451,7 @@ p1_targets_list <- list(
     p1_drb_distance_matrix_updown,
     p1_drb_distance_matrix$updown
   ),
-  #save as csv for use in Python
+  #save as csv for readability
```
lekoenig, line 454: I was wondering whether we still want to keep this csv file target, but I guess it doesn't hurt to keep as a way of quickly checking the format/dimensions of the updown matrix.

1_fetch.R
```diff
@@ -462,6 +463,14 @@ p1_targets_list <- list(
     },
     format = 'file', 
     repository = 'local'
+  ),
+  #save as npz for use in Python
```
lekoenig, line 467: Thanks for adding the npz target! Do any of our other inputs need to be formatted as npz, or just the distance matrix?

jds485: Just the distance matrix. The other inputs are attributes and SC observations, which are both saved as zarr files.

2_process/src/write_data.R
jds485: Thanks! I meant to add that comment. I could not figure out how to get the `keys` automatically recognized within `savez_compressed`. In Python, a named dict is automatically converted to have the same names in the npz file, but that did not happen using reticulate

jds485:
@lekoenig I think I addressed your comments. Let me know if this is good to merge. I approved based on your edits

jds485 (approved):

1_fetch.R
lekoenig: Sounds good, thanks for the reminder about the expected file formats!

lekoenig:
Looks great, @jds485! I'll defer to you to merge whenever you think it's ready. I tried to "approve" the PR, but GitHub apparently won't let me do that on my own PR 😄

PR #203, opened 2022-11-09
Remove random selection of attributes

### LINKED ISSUE #200: Remove random attribute selection
Currently, highly correlated attributes are removed by randomly selecting 1 set of correlated features and removing all but 1 of those features. I had a similar selection method in another [repo](https://github.com/USGS-R/regional-hydrologic-forcings-ml/pull/156#pullrequestreview-1170723882), and Charlie pointed out this method can make reproducibility difficult (we'd need to provide the targets random seed to reproduce what we currently select). This issue is to remove the random selection to a selection based on feature order.

jds485:
This PR removes the random selection of attributes in correlation screening to make the results simpler to reproduce. I also added print statements as the attributes are being removed.

The meta file is up to date for all except `p4` targets. I'll remake those targets in a separate PR.

Closes #200


galengorski (approved):
@jds485 This looks good to me! I ran it twice on tallgrass and it produced the same set of features. Fine by me to merge

PR #207, opened 2022-11-09
Updated functions to handle spatial splits

### LINKED ISSUE #205: Add spatial holdout option
After #191, we can now implement spatial holdouts (reaches) using a sampling design that provides the same proportion of reaches with discrete and continuous data in each data split (training vs. testing, analysis vs. assessment)

jds485:
This PR adds options to split data by spatial information (reaches in this case). It is applied to the training and testing split, as well as to CV folds in parameter tuning. The main functions for this are in train_models.R: `make_spatial_split`, `make_spatial_split_CVtraining`, and `assign_spatial_split`. The first and second functions are similar to the temporal split function. The last function is used to ensure consistency in the training and testing sets across all models that will be compared.

I set the random and temporal targets to never rebuild because there were function edits that would trigger them to rebuild.

Additional visualization edits:
- visualizations for random, spatial, and temporal split results in 3_visualize.R
- adds DRB boundary to all maps to better see where the estuary is located
- adds manual colors to some plots in plot_dagnostics.R to ensure consistency across reaches
- adds shapviz graphics for SHAP plots. These are updated for every model in 4_predict_plots.R. I had an error message for SHAP computations for `static_dynamic_spatial` that I'd not seen before. I commented out and am trying to diagnose the issue now.

Closes #205


galengorski (approved):
This looks good to me @jds485 , I mainly focused on the functions within `train_models.R`. One thought for spatial validation that might be informative (and potentially more complex) is to use subwatersheds or subregions to do stratified sampling. Depending on how site specific you think the dynamics are, it could be an interesting way to discover regional generalizability. Just a thought

jds485:
Thanks! Yes, accounting for spatial nestedness is on my list. I like the suggestion to define sub-watersheds or regions to hold out. One of the challenges I've been thinking about is how to ensure a proportional amount of discrete and continuous sampling data within each sub-region. Maybe we can just accept that some sub-regions will have different amounts (and that is likely for prediction in other basins, anyway).

PR #208, opened 2022-11-10
Update p4 targets after recent PRs

### LINKED ISSUE #204: Update p4 targets after recent PRs
I've been holding off on updating the `p4` targets for RF models because I knew there would be several PRs that affect results. After #203 is merged, the `p4` targets can be updated.

jds485:
This PR updates our `4_....R` scripts and `p4` targets to match edits made in recent PRs.

Summary of edits:
- removed the `cue = tar_cue('never')` from all targets
- retained the new `data_type` variable and made sure it's not used as a predictor
- updated the static-only model results so that the `#Correct workflow problem` workarounds could be removed
- For SHAP values, 25% of the data are now used to reduce RAM and time requirements. For the most important features, their ranks are similar using the random 25% sample and full dataset.
- Adds a more complete grid search for the best performing RF model (`p4_train_RF_static_dynamic_temporal_fulltune`)

For some reason, all of `p4_predict_plots.R` is appearing as changed. I only edited the `p4_shap` targets that compute SHAP values starting with the comment `SHAP values and plots`. SHAP plot targets were not changed.

Closes #204


jds485:
Self-merging after discussion in weekly meeting.

Issue #189, opened 2022-10-21
Add function to plot empirical CDFs of model performance


jds485:
function should accept a list of model results as input and make one curve per model.

Model results columns 
residuals
PRMS_segid
Date

Example targets with model predictions:
`p4_pred_RF_static_dynamic_temporal_test`
`p4_pred_RF_min_static_dynamic_temporal_test`
`p4_pred_RF_dynamic_temporal_test`
These are lists. The `$preds` list is a data frame with errsq (squared residual error for each day), PRMS_segid, Month, and Year (and some other columns).

Could have options to make:
a CDF plot for each reach
one CDF over all reaches
a CDF for selected months/years (facet wrapped?)

Example: summary RMSE by reach:
summarize(group_by(p4_pred_RF_static_dynamic_temporal$pred, PRMS_segid), 
                                 RMSE = sqrt(mean(errsq)),
                                 RMSE_log10 = log10(RMSE))

Axes: 
empircal CDF quantiles
error metric values for those quantiles (e.g., RMSE)

jds485:
@lekoenig: I think this would be most helpful to have for AGU

lekoenig:
Thanks for providing all of these ideas, @jds485! Here are some plots from a function that can either plot:

1) one CDF over all reaches:
![test](https://user-images.githubusercontent.com/8785034/205694905-f2464499-7776-41fe-a00a-7ef5813f1b7a.png)


2) a CDF plot showing the CDF of RMSE's for each date for any selected reach(es):
![test2](https://user-images.githubusercontent.com/8785034/205694939-f9af096a-143e-414e-8bba-963156a69fe8.png)


3) a plot that shows the CDF over all reaches, but only for selected months and years:
![test3](https://user-images.githubusercontent.com/8785034/205694961-ecb615eb-9aae-49bc-87fb-80261f6daf29.png)

The user can choose whether to show a logged x axis or not. We can edit or parameterize any of the plot formatting, but let me know if these plots are what we're looking for and I can open a new PR that addresses this issue.

jds485:
Thanks! These look good to me. Do you have an option to plot results for several models (one CDF per model)?

lekoenig:
> Do you have an option to plot results for several models (one CDF per model)?

[...]

PR #215, opened 2022-12-05
Add function to plot empirical cdf

lekoenig:
This PR adds a function, `plot_ecdf`, to plot empirical CDFs of model performance given a data frame of model results. The plot includes 1 CDF curve for each model represented in the model results. I opted to format the model results as a data frame instead of a list, which we discussed in #189.  I think it makes the function more generalizable to move data formatting code outside of the plotting function. The model results list is formatted as a data frame in `p3_model_results`.

Much of the plot formatting is already parameterized, but we could other formatting options like color more flexible. Here's what `p3_ecdf_all_reaches_png` looks like for me:

ecdf_all_reaches

### DIFF
--- a/3_visualize/src/plot_ecdf.R
+++ b/3_visualize/src/plot_ecdf.R
@@ -1,0 +1,136 @@
+#' @title Plot empirical CDFs of model performance
+#' 
+#' @description
+#' Function to plot empirical CDFs of model performance and make one
+#' curve per model. The user can adjust the plot aesthetics using 
+#' several arguments that control plot formatting.
+#' 
+#' @param model_results data frame containing the model results. Must contain
+#' columns "model", "PRMS_segid", "Date", and "errsq".
+#' @param plot_type character string indicating the type of eCDF plot to return.
+#' Options include 1) "all_reaches" to plot one CDF over all reaches; 
+#' 2) "select_reaches" to return one CDF for each reach within `segid_select`; and
+#' 3) "select_times" to return a grid of CDF plots for each month and year in 
+#' `months_select` and `years_select`, respectively. Defaults to "all_reaches".
+#' @param fileout character string with the name of the saved cdf plot, including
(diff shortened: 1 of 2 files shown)


jds485:
Looks good! Is the code ready for review? / Do you think it needs a review?

lekoenig:
> Looks good! Is the code ready for review? / Do you think it needs a review?

Thanks, this is ready for review! I don't think this needs an extensive review so long as it meets the proposed use cases. I'd appreciate it if you could check that the RMSE values I'm using to plot the ecdf match what you were expecting (e.g. summarization steps in lines 45-50; 67-72; and 90-98 of the `plot_ecdf.R` script).

3_visualize/src/plot_ecdf.R
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+
+  # To plot one CDF plot for each reach, compute the RMSE for each date
+  # for each reach, and then make the CDF for those values. (This procedure
+  # is equal to taking the sqrt of errsq for each row of model_results$pred
+  # since each segment has only one error value for every date.)
+  if(plot_type == "select_reaches"){
+    preds <- model_results %>%
+      filter(PRMS_segid %in% segids_select) %>%
```
jds485, line 68: does this make 1 CDF per reach in `segids_select`? Or would we branch over all segids to get a plot for each one?

jds485 (approved):
Thanks, Lauren! This looks good. I just have 1 question for you

3_visualize/src/plot_ecdf.R
lekoenig: If more than one segment is requested, the function will return a faceted plot where each segment gets its own panel, like this: 
test_reaches

This probably gets unwieldy when trying to visualize all of the reaches, so keep me updated with how you're using this function.

PR #219, opened 2023-01-25
Minor SHAP plot edits

jds485:
This PR presents minor edits to SHAP plot visualization functions. Arguments are added to specify x-axis limits or a scaling factor (exponent) to better highlight the SHAP values that are closer to 0 instead of the extremes. I've noted in the function that using the scaling factor can result in a different ranking due to the non-linear scaling. I decided to not use the scale factor and instead uses different x-limits.

Resulting plot looks like this:
image

Previously, based on x-axis data range, plots looked like this:
image


lekoenig (approved):
Looks good to me 👍

Issue #216, opened 2022-12-07
Add timer for when to refresh AWS credentials


jds485:
The targets that refresh AWS credentials take a while to run when all dependency targets have been built. ~20 AWS refreshes in a row takes ~30 mins. I think we should add a timer that is used within each of the AWS refresh targets to indicate if the credentials should be refreshed or not.

jds485:
Update: The time is largely spent pulling the `dummy_var` from S3. Generating the credentials file is quick. As far as I can tell, targets will load all of the dependency targets before executing the command. So, I'm not sure how to speed up these AWS refresh targets. I think in the future it will be best to include the generate_credentials() function at the end of every long-to-run target / function.

PR #221, opened 2023-02-03
220 seasonal shap

### LINKED ISSUE #220: Add function(s) for seasonal feature importance analysis
We would like to understand which features are being used to predict wintertime pulse SC event after snow / road salt is applied, and also which features are used to predict during summer low flow conditions.

- split data into water year seasons
- compute SHAP values for each of those seasons
- For low flows (summer), also split the seasonal data into watershed regions with distinct baseline geochemistry (physiographic regions)
- compute SHAP values for each of those seasons and physio regions
- Visualize results in panel plots
- For winter, also split the seasonal data by 1) high % urban and 2) high % forest
- compute SHAP values for each of those land cover splits

jds485:
This PR provides functions that split our dataset into:
- water year seasons
- land cover classes (high urban, high forest cover)
- physiographic province
- water year seasons by land cover class
- water year seasons by physiographic province

1_fetch additions:
- fetches the physiographic ecoregions from a SB item.
- Adds the unique region names to each PRMS reach. This was a bit of a pain because reaches overlap multiple regions, and some regions were split into several polygons. I'm curious if you know of a better way to extract the ecoregion data than what I programmed.

2_process additions:
- adds a target `p2_TOT_lc_physio_attrs` that contains only the TOT land cover classes and physiographic identifiers (binary for each region) for each SC observation.

4_predict additions:
- adds a function `setup_shap_data` that completes the dataset splits and returns a list with one element per split. The target that uses this function is `p4_shap_data_splits` in 4_predict_plots.R.
- adds a function `get_shap_subset` that filters the full dataframe of SHAP values for only the rows in each data split
- adds a function `get_shap_dir` that gets the output file directory based on the data split name
- set the Boruta and spatial train/test targets to never rebuild. These were triggered to rebuild because of the edits to `generate_credentials`

_targets.R additions:
- adds directories for the SHAP analyses corresponding to the data splits

I also addressed #216, but do not have a solution to the problem. I added a function in `generate_credentials.R` that checks if the most recent credentials file was generated in the last 10 minutes. As noted in #216, the majority of the time is spent loading the `dummy_var` target that is used to trigger when the aws credentials should be built. As part of this process, I realized that many of the aws credentials targets could be deleted because the last line in the dependency target's function calls `generate_credentials()`. That helped to reduce the overall time spent on these aws credentials steps. I recommend that in the future we regenerate credentials as part of functions instead of as separate targets.

Closes #220, #216


jds485:
Here are example results from Urban and Forest. There are clear differences in feature importance rank, and also direction for some attributes (inverse pattern for TOT Clay).

Urban (TOT > 75%)
SHAP_global_RF_static_dynamic_temporal_full_vars40

Forest (TOT > 75%)
SHAP_global_RF_static_dynamic_temporal_full_vars40

jds485:
I'm curious what reaches are in each of the spatial data splits (land cover, physio region). I can make a map target for these splits. Would you prefer as a separate issue?

lekoenig:
> I'm curious what reaches are in each of the spatial data splits (land cover, physio region). I can make a map target for these splits. Would you prefer as a separate issue?

Unless you have a preference, I think that'd be helpful so that I can focus on the code changes in this PR.

jds485:
Okay, I'll make a separate issue to map these

1_fetch/src/check_physio_region.R
```diff
@@ -0,0 +1,56 @@
...
+  physio_mat <- data.frame(matrix(nrow = nrow(reaches), 
+                                  ncol = length(unique(ecoregions$Physio))))
+  colnames(physio_mat) <- sort(unique(ecoregions$Physio))
+  for(i in 1:ncol(physio_mat)){
+    #several polygons define each ecoregion, so using a function to check
+    # all of those polygons for each reach
+    physio_mat[,i] <- apply(physio_intersect, MARGIN = 1, 
+                            FUN = check_physio_region, region_name_check = colnames(physio_mat)[i],
+                            region_names = ecoregions$Physio) 
+  }
```
lekoenig, line 27: > I'm curious if you know of a better way to extract the ecoregion data than what I programmed.

Woah, this approach is really clever! I don't know if I have a better way to extract the ecoregion data - do you want to know all of the ecoregions a line intersects, or just the one that overlaps the most? 

For spatial joins, I like the flexibility of `st_join(x, y, join = st_nearest_feature)` because you can choose a series of join operations. For example, I can specify `join = st_intersects` or `join = st_nearest_feature`. I'm guessing that's not so simple with this dataset though - `st_intersects` would return multiple lines if the reach crosses ecoregion boundaries; `st_nearest_feature` would (I think) find the ecoregion closest to one of the vertices so that ecoregion might not be the one that overlaps the most. 

If looking for the region that overlaps the most, I probably would approach it by finding the region with the greatest intersecting length:
```r
physio_intersections <- st_intersection(x = reaches, y = ecoregions) %>% 
    mutate(len = st_length(.)) %>%
    group_by(subsegid) %>%
    arrange(desc(len), .by_group = TRUE) %>%
    slice(1) %>%
    ungroup() %>%
    st_drop_geometry() %>%
    select(subsegid, Physio) 
  
reaches_w_regions <- left_join(x = reaches, y = physio_intersections, by = "subsegid")
head(reaches_w_regions, 2)
#> Simple feature collection with 2 features and 8 fields
#> Geometry type: LINESTRING
#> Dimension:     XY
#> Bounding box:  xmin: -74.61166 ymin: 42.14689 xmax: -74.46774 ymax: 42.356
#> Geodetic CRS:  WGS 84
#>   subsegid subsegseg subseglen fromsegs toseg tosubseg segidnat Physio                       geometry
#> 1      1_1         1  13588.58     <NA>     3      3_1     1435     AP LINESTRING (-74.51816 42.35...
#> 2      2_1         2  19134.73     <NA>     8      8_2     1436     AP LINESTRING (-74.47273 42.23...

```

That doesn't give you the binary matrix though. I tried modifying this quickly to keep all intersecting regions if that's what you're looking for. Hopefully these ideas are helpful for showing other ways to approach this problem. I haven't checked either of them for accuracy or against your original values. 

```r
physio_intersections <- st_intersection(x = reaches, y = ecoregions) %>%
  st_drop_geometry() %>%
  select(subsegid, Physio) %>%
  distinct() %>%
  pivot_wider(names_from = Physio, values_from = Physio) 
  
reaches_w_regions <- left_join(x = reaches, y = physio_intersections, by = "subsegid") %>%
  mutate(across(BR:VR, ~if_else(is.na(.), 0, 1)))

head(reaches_w_regions, 2)
#> Simple feature collection with 2 features and 12 fields
#> Geometry type: LINESTRING
#> Dimension:     XY
#> Bounding box:  xmin: -74.61166 ymin: 42.14689 xmax: -74.46774 ymax: 42.356
#> Geodetic CRS:  WGS 84
#>   subsegid subsegseg subseglen fromsegs toseg tosubseg segidnat BR AP CP PD VR                       geometry
#> 1      1_1         1  13588.58     <NA>     3      3_1     1435  0  1  0  0  0 LINESTRING (-74.51816 42.35...
#> 2      2_1         2  19134.73     <NA>     8      8_2     1436  1  1  0  0  0 LINESTRING (-74.47273 42.23...
```

4_predict.R
```diff
@@ -95,15 +95,8 @@ p4_targets_list <- list(
...
-  ),
-  
-  # Refresh AWS credentials
-  tar_target(
-    p4_aws_credentials_1,
-    generate_credentials(dummy_var = p4_Boruta_static),
-    deployment = 'main',
-    cue = tar_cue('always')
+             deployment = 'worker',
+             cue = tar_cue(mode = "never")
```
lekoenig, line 99: I understand why you've taken this approach, but `tar_cue(mode = "never")` does make me a little wary in that it puts onus on the user to understand when parts of the pipeline should be updated or not. Could we add notes somewhere in the documentation about the use of `tar_cue(mode = "never")`? Or do we plan to eventually take these options out and re-run the pipeline in full? Not necessary to address for this PR, just something I wonder about.

4_predict/src/train_models.R
```diff
@@ -555,6 +555,228 @@ filter_rows_date <- function(attrs, start_date){
...
+  #' 
+  #' @return Returns a list with the different data splits
+  
+  data_lst <- list()
+  data_lst_names <- c()
+  
+  if(seasonal){
+    #split data by water year season (4 elements in list)
+    data$months <- lubridate::month(data$Date)
+    OND <- filter(data, months %in% c(10,11,12)) %>%
```
lekoenig, line 581: lubridate also has a function called `quarter` that could be useful for cases like this (for future reference, this looks good and seems like you're actually using these list names).

4_predict/src/train_models.R
```diff
@@ -555,6 +555,228 @@ filter_rows_date <- function(attrs, start_date){
...
+      select(PRMS_segid, Date)
+    
+    data_lst <- c(data_lst, list(OND), list(JFM), list(AMJ), list(JAS))
+    #names for list elements
+    data_lst_names <- c(data_lst_names, paste0('seas_', c('OND', 'JFM', 'AMJ', 'JAS')))
+    
+    if(lulc){
+      #split each season by lulc
+      ind_seasons <- grep('^seas_', data_lst_names)
+      for(i in 1:4){
```
lekoenig, line 597: To make the for loops more robust to unexpected values of `ind_seasons` you could consider adding a check to this function that the length of `ind_seasons` is 4, e.g. `stopifnot(length(ind_seasons) == 4)` or just loop by `seq(ind_seasons)`.   This is a minor suggestion though, so I'll leave that up to you.

4_predict/src/train_models.R
```diff
@@ -555,6 +555,228 @@ filter_rows_date <- function(attrs, start_date){
 
 
 #SHAP values
+setup_shap_data <- function(data, seasonal = TRUE, physio = TRUE, lulc = TRUE,
+                            lulc_prop = 0.75, lulc_data = NULL, physio_data = NULL){
+  #' @description prepares data for SHAP computation based on the provided split
+  #'
+  #' @param data the dataset with predictor attributes
+  #' @param seasonal logical for splitting data by water year season
```
lekoenig, line 563: I found these argument descriptions helpful, thanks! The arguments themselves might be more intuitive by rewording slightly, e.g. `split_by_season` instead of `seasonal`. This is a very minor suggestion and doesn't influence the outcome of the code, so I leave this up to you.

4_predict/src/train_models.R
```diff
@@ -555,6 +555,228 @@ filter_rows_date <- function(attrs, start_date){
...
+  sample <- shap[shap$ind %in% sample_inds$ind, 
+                 -which(colnames(shap) == 'ind')]
+  
+  #add the name of the list to this and create a list
+  lst <- c(list(name = split_name), list(shap = sample))
+  
+  return(lst)
+}
+
+get_shap_dir <- function(main_dir, subdir_name){
```
lekoenig, line 713: Woah, nice job handling these directories 🙂

lekoenig (approved):
Looks good to me, @jds485! I did not run the pipeline but I checked the code changes and tried to briefly inspect the relevant targets to get a sense for what each is doing. These changes will be really helpful for interpreting the RF model output. I left a few minor questions or suggestions. Since these do not impact the output or efficiency of the code, feel free to take or leave whatever is helpful for you.

One question - I did not fully follow the credentials issue. Why are the intermediate credentials targets omitted here? Because you discovered generating these targets is not the time-limiting step?

jds485:
Thanks, Lauren! Good call on not running the target computations. I forgot to add time requirements in my initial post. The 3 full-dataset SHAP computations take a total of about 15 hrs to run + another 4 to make plots and summarize results.

> Why are the intermediate credentials targets omitted here? Because you discovered generating these targets is not the time-limiting step?

The credentials targets that I deleted were all redundant because they were dependent on targets whose functions have a `generate_credentials()` call at the end.

_targets.R
```diff
@@ -51,18 +51,29 @@ train_test_features <- c("RF_static", "RF_min_static", "RF_static_dynamic",
                          "RF_min_static_dynamic","RF_dynamic")
 train_test_res <- c("pred_obs","spatial_res","monthly_res","annual_res","temporal_res")
 rf_xai_plot_types <- c("shap","dependence") 
+rf_xai_shap_options <- c("seasonal","lulc", "physio")
+rf_xai_shap_lulc_options <- c("high_forest", "high_urban")
+rf_xai_shap_physio_options <- c("appalachian", "coastal", "interior")
+rf_xai_shap_seasonal_options <- c("OND", "JFM", "AMJ", "JAS")
+rf_xai_shap_seasonal_suboptions <- c(rf_xai_shap_lulc_options, rf_xai_shap_physio_options)
```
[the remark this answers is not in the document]
jds485: Most of the created directories are not used, so I'm happy to add an issue to add a directory deletion function / code.

I think all of the temporal and spatial train/test RF models will contribute to the paper (main text or supplement). The random split likely will not unless a reviewer asks (unlikely). I am not planning to further edit the features used. Let's keep all of these targets for now, but I'll add a note to revisit for the data release/model archive.

I'm debating how is best to store the river-dl results. I'm currently developing in snakemake, and that can be a standalone workflow. I'm curious on your thoughts about time required to integrate with targets (I think DO quest did that?)

1_fetch.R
```diff
@@ -471,6 +472,47 @@ p1_targets_list <- list(
...
+            exdir = "1_fetch/out", overwrite = TRUE)
+      paste0("1_fetch/out/", 
+             unzip(zipfile = grep("physiographic_regions_DRB",p1_drb_ecoreg_zip,value=TRUE), 
+            exdir = "1_fetch/out", overwrite = TRUE, list = TRUE)$Name
+            )
+    },
+    format = "file",
+    deployment = 'main',
+    repository = "local"
+  ),
```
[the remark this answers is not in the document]
jds485: Yes, I'm sure that will work! Let me try it out before committing

_targets.R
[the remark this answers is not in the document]
lekoenig: > I'm debating how is best to store the river-dl results. I'm currently developing in snakemake, and that can be a standalone workflow. I'm curious on your thoughts about time required to integrate with targets (I think DO quest did that?)

Yes, we did integrate the targets data prep pipeline and the snakemake modeling pipeline for the DO quest. We do that by implementing system calls to kick off the snakemake workflow within a [target](https://github.com/USGS-R/drb-do-ml/blob/main/2a_model.R#L209L244) that builds model metrics. However, this adds complexity to the workflow and in my opinion makes it more difficult to spot or diagnose errors. Unless there's a good reason to integrate the pipelines, in the future I will more likely make the workflow more modular and keep the targets and snakemake pieces separate.

1_fetch/src/check_physio_region.R
jds485: Thanks! This is helpful to see other approaches.

> do you want to know all of the ecoregions a line intersects, or just the one that overlaps the most?

Ideally, I'd want to know the ecoregion with the most drainage area overlap for the PRMS catchment, so instead of a binary matrix it would be a matrix of proportion drainage area within each region. I decided to keep it simple based on reach intersection instead of PRMS catchment intersection for time.

_targets.R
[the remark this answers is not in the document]
jds485: Okay, thanks! I'll continue to develop the river-dl outside of `targets`

4_predict.R
jds485: Yes, I'll edit the README to say which targets are currently set to never update. The critical target to never update for the current work is the `p4_Boruta_static` because that determines which features are used in the models. The model targets could all have the `tar_cue(mode = "never")` removed (it will change model results, but hopefully not by much). I can add an issue to update the pipeline to remove the never cue for model targets. I currently see this as lower priority that can be addressed after the RGCN results are ready.

lekoenig: Thanks, Jared. I agree that this is a lower priority and am happy to go with whatever you think is best. I want to avoid a situation where a modeling target doesn't update when we expect it to due to this setting of `tar_cue` but I'm also less familiar with the specifics of the modeling steps and tradeoffs re. computational efficiency.

4_predict/src/train_models.R
jds485: As programmed, I think `data_lst_names` is guaranteed to have 4 elements. Some of the elements might be a data frame with 0 rows, but the element will exist. But yeah, it's possible that we later edit the function to rearrange the order of these function's steps and then this loop could give unexpected results. I'll change this

PR #225, opened 2023-02-17
added functions to plot the data splits

### LINKED ISSUE #222: Create maps of XAI data splits
Create target(s) that map the reaches within each of the XAI spatial data splits (land cover, physio region).

Physio region can be the polygon outlines
Land cover by color (red = urban, green = forest, all others = gray)

jds485:
Added a function to plot the XAI data splits. Here are some examples for the urban and forest split. Gray are reaches with no data, dark blue are reaches with data in the titled split, and light blue are reaches with data but not in the split. Suggestions for how to get this info into a legend on the plot are welcome, but it's low priority.

Based on this map, I may want to lower the urban threshold. The 75% threshold is only met around Philadelphia.
lc_highUrban

lc_highForest

Turns out that it was also useful to map the seasonal splits. We lose some spatial coverage in the JFM winter months. 

seas_JFM

Closes #222


4_predict_plots.R
```diff
@@ -1970,6 +1970,20 @@ p4_plot_targets_list <- list(
                       select(PRMS_segid, Date, AP, BR, VR, CP, PD)
                     )
   ),
+  #Make maps showing the reaches that each data split covers
+  tar_target(
+    p4_shap_data_split_maps_png,
+    plot_data_splits(split = p4_shap_data_splits[[1]], 
+                     split_name = names(p4_shap_data_splits)[1],
+                     #using p2_TOT_lc_physio_attrs because it's smaller than the full table
+                     full_df = full_df <- distinct(p2_TOT_lc_physio_attrs, PRMS_segid),
```
lekoenig, line 1979: Suggesting this change in case of a typo. 

And is `full_df` expected to contain columns other than `PRMS_segid`? I see that `plot_data_splits` has another call to `distinct(full_df, PRMS_segid)` so just want to make sure I'm understanding the inputs.

```suggestion
                     full_df = distinct(p2_TOT_lc_physio_attrs, PRMS_segid),
```

3_visualize/src/plot_nhdv2_attr.R
```diff
@@ -71,3 +71,55 @@ plot_nhdv2_attr <- function(attr_data,network_geometry,file_path,
...
+              sf::st_as_sf(),
+            size = 0.3, 
+            color = 'lightblue') +
+    #split data
+    geom_sf(data = split %>% 
+              left_join(.,network_geometry[,c("subsegid","geometry")],
+                        by=c("PRMS_segid"="subsegid")) %>%
+              sf::st_as_sf(),
+            size = 0.3, 
+            color = 'blue') +
```
lekoenig, line 113: The easiest way to add the colors to a legend is to map the colors onto a plot aesthetic, i.e., by using `aes()` within the layered graphics and then defining the colors using `scale_color_manual`. I think this should give the correct legend, although the plot dimensions probably get wonky: 

```suggestion
    geom_sf(data = network_geometry, 
            aes(color = "no data"),
            size = 0.3) +
    #DRB boundary
    geom_sf(data = boundary, fill = NA,
            size = 0.3, color = 'black') +
    #full data
    geom_sf(data = full_df %>% 
              left_join(.,network_geometry[,c("subsegid","geometry")],
                        by=c("PRMS_segid"="subsegid")) %>%
              sf::st_as_sf(),
            aes(color = "data present in full set"),
            size = 0.3) +
    #split data
    geom_sf(data = split %>% 
              left_join(.,network_geometry[,c("subsegid","geometry")],
                        by=c("PRMS_segid"="subsegid")) %>%
              sf::st_as_sf(),
            aes(color = "data present in split set"),
            size = 0.3) +
    scale_color_manual(name = "",
                       labels = c("no data", "data present in full set", "data present in split set"),
                       values = c("gray", "lightblue", "blue")) + 
```

lekoenig (approved):
Looks good to me, @jds485! I made a suggestion for how to add a legend, hopefully that works for this example. 

I also had one (maybe unrelated) question - since `drb_shp` is defined in `_targets.R` will  `p4_shap_data_split_maps_png` always show as outdated? Or perhaps it always shows up as up-to-date even if the input huc 8's change? No changes necessary, I'm just curious about the different behaviors with building the shapefile in `_targets.R` versus defining a separate target.

4_predict_plots.R
jds485: Thanks! That was a typo.

`full_df` only needs the `PRMS_segid`

3_visualize/src/plot_nhdv2_attr.R
jds485: Thanks! This mostly worked, but the layer order on the maps wasn't correct, and it seemed to always plot the "full set" first, regardless of the position in the labels vector. Is there a way to specify the layer order?

jds485:
>  since drb_shp is defined in _targets.R will p4_shap_data_split_maps_png always show as outdated? Or perhaps it always shows up as up-to-date even if the input huc 8's change? 

`p4_shap_data_split_maps_png` does not show as outdated with the files as-is. Even though `drb_shp` is not a target, it is still tracked in the meta file:
drb_huc8s|object|3fa2976b525e0553|||||||||||||||
drb_shp|object|4e7d7d86f5133785|||||||||||||||

So, I think changes to this file will trigger `p4_shap_data_split_maps_png` to be outdated.

3_visualize/src/plot_nhdv2_attr.R
lekoenig: Oh, right - it plots the layers by alphabetical order. I think in the past I've used the `breaks` argument to make sure the color `values` align with the layers, but if your approach works as you expect I think that's as good an option as any. Good to know!

PR #226, opened 2023-02-21
Remove cue of never and do full grid search

### LINKED ISSUE #224: Update the pipeline to remove the "never" cue for model targets.
RF model targets were set to never update after the edit to `generate_credentials()` in #221. These targets can be rebuilt at the expense of a few days of computation. For the temporal and spatial train/test split, it would help to increase the number of grid points evaluated for the final models when these are updated. The random split does not need to have more grid points.

jds485:
This PR removes the tar_cue('never') from model training targets and updates the meta file with these changes.

For the temporal and spatial train/test split models, I also increased the number of cross validations and number of grid points to use within the hyperparameter tuning. For the spatial and temporal splits, I deleted a target whose name ended with `_fulltune` because the hyperparameter settings are the same as the new settings. I did not increase the CV folds and grid points for the random train/test models because they will not be used in the manuscript.

I discovered that some models needed to be trained with fewer cores (40) and some could be trained with more (65) depending on the RAM use. Turns out that tidymodels must prevent crashes that would be caused by exceeding the RAM limit, and instead proceeds slowly to be within the RAM limit. For example, a model trained on 50 cores did not finish in 8 hrs, but training the same model on 40 cores finished in 1 hr. This is something to be aware of for models that are RAM intensive.

I also discovered that the SHAP calculations are sometimes slowed down due to the RF number of threads parameter. I set this to 1 thread for some models that ran slowly.

Closes #224


jds485:
Best overall RMSE spatial train/test model - R2_trad = NSE < 0, so not a good model!
hyperparam_diagnostic_daily_SC_RF_min_static_dynamic_spatial

Best overall RMSE temporal train/test model:
hyperparam_diagnostic_daily_SC_RF_static_dynamic_temporal

galengorski (commented):
This looks good to me @jds485 . Just so I understand, the reason for removing `tar_cue` was to facilitate the hyperparameter grid search, correct? Do you plan on putting `tar_cue` back in once you're done hyperparameter tuning?

jds485:
Yes, I needed to remove `tar_cue(cue = 'never')` so that those targets would update with the new grid search settings. I originally used `tar_cue(cue = 'never')` for model targets because I made a small change to the training function that did not affect calculations (AWS credential generation), but it would have triggered all of those models to rebuild. Hopefully no more changes like that will be needed, so I'm not planning to add the `tar_cue(cue = 'never')` back to these targets

galengorski:
Got it ok, that makes sense, do you have (or think you'll need to have) safe guards against triggering a full model rebuild on accident?

jds485:
> do you have (or think you'll need to have) safe guards against triggering a full model rebuild on accident?

There are no safeguards against triggering model rebuilds, but if they are triggered and build, then we can easily `git stash` or `git revert` back to previous versions if we need to. The targets meta file is committed to the repo and links to the S3 data. I've used `git stash` a few times while developing figures to test a function interactively, git stash, then uninteractively run it and make sure results are the same

Issue #192, opened 2022-10-28
Target branching over RF models


jds485:
To reduce code length, it would be nice to implement targets branching for RF model targets. Probably best to implement for the plots in `4_predict_plots.R` instead of the model training targets in `4_predict.R` because individual models have different specifications, whereas I think all models have the same plot function args.

Idea from @lekoenig [here](https://github.com/USGS-R/drb-inland-salinity-ml/pull/187#discussion_r1008355167): 
> I wonder if, in a new target, we could add elements to those lists that contain the model_name and out_dir and then map over each list to populate the arguments in plot_vip.

PR #229, opened 2023-03-15
added plots for RGCN model results

jds485:
This PR adds targets that plot results for the RGCN models. Plots are the same as we plot for RF, aside from SHAP plots that are not ready for RGCN models yet.

Note: click the hide whitespace button. Line endings must be making entire files appear different.
image


jds485:
Quest team discussed this PR in our meeting last week. Decided it is good to merge, but should use target branching for plots, as documented in #192.

PR #232, opened 2023-04-07
Prepare for Upcoming DGEC Transition

jesse-ross:
Starting on September 30, 2023, USGS GitHub repositories will no longer be permitted to remain in existing organizations such as USGS-R on the public GitHub. Instead, they must be either (a) migrated to the [DOI GitHub Enterprise Cloud (DGEC)](https://github.com/DOI-USGS/), (b) migrated to the [USGS GitLab instance](https://code.usgs.gov/), or (c) archived.

In order to be made public, all USGS software, whether [provisional or approved](https://www.usgs.gov/products/software/software-management/types-software-review), must contain all necessary legal and metadata assets. The purpose of this PR is to ease the transition by creating the necessary files, so that people can do the necessary migration work with less research time.

The files added here are generated based on some assumptions, and without looking at the contents of any existing files. For instance, if this repo already has a `LICENSE.md`, the process which is generating this PR will not add one to the PR, but it does not make any attempt to validate the existing file. Similarly, the code is not at all clever about finding similarly-named files. For instance, if you already have a `LICENSE.txt`, this PR will try to create a `LICENSE.md` anyways!

The `DISCLAIMER.md` provided here is for a provisional release. If an official release is intended, a different disclaimer template may be found [here](https://code.chs.usgs.gov/software/software-management/-/tree/main/administrative_templates)

The `LICENSE.md` provided here is the most common choice, but as the owner of the repository you need to look it over and make sure that it is a good match for your repository. The [Office of the Solicitor](https://www.doi.gov/solicitor) is the only entity that can provide legal advice on this matter.

 Note especially that the `code.json` created by this PR will need to be manually edited in order to be correct. In this file, the following are likely to require modification:
* the "version" element simply chooses the repository's default branch, e.g. `main`. If there are tagged versions, they will need to be added by hand.
* the "status" element is listed as "Development". A correct value should be chosen from these possibilities: "Ideation", "Development", "Alpha", "Beta", "Release Candidate", "Production", or "Archival"
* the "hours" element is listed as -1. If an actual estimate of hours can be put in here, it should.
* the "tags" element only has "USGS". Other tags may be desirable to be added.
* the contact name is filled in with the GitHub username of the user with the most commits to the repo. It should be replaced with an actual name.
* the contact email must be filled in manually.
* the "metadata_last_updated" element should be filled with the date when you make the above changes.
* the `code.json` file will contain two sections, which are intended for different branches or version releases. You may want to remove one of these, or you may want to add more.

For examples of valid `code.json` files, please see [https://github.com/DOI-USGS/dataRetrieval/blob/main/code.json](https://github.com/DOI-USGS/dataRetrieval/blob/main/code.json) or [https://github.com/DOI-USGS/snow-to-flow/blob/main/code.json](https://github.com/DOI-USGS/snow-to-flow/blob/main/code.json)In addition to the necessary legal and metadata assets, all public software requires a security review. See [this guidance](https://www.usgs.gov/products/software/software-management/types-software-review) for more information. Additionally, a useful FAQ with links to details of USGS policy and the specifics of this migration is located [here](https://doimspp.sharepoint.com/sites/Software/SitePages/Frequently-Asked-Questions.aspx?xsdata=MDV8MDF8fDljYTRhZTEwNzE3MjRmYmU5YjYzMDhkYWQ5ODhkMjg3fDA2OTNiNWJhNGIxODRkN2I5MzQxZjMyZjQwMGE1NDk0fDB8MHw2MzgwNjE0Nzg4OTcyMjg2NDN8VW5rbm93bnxWR1ZoYlhOVFpXTjFjbWwwZVZObGNuWnBZMlY4ZXlKV0lqb2lNQzR3TGpBd01EQWlMQ0pRSWpvaVYybHVNeklpTENKQlRpSTZJazkwYUdWeUlpd2lWMVFpT2pFeGZRPT18MXxNVFkzTURVMU1UQTRPRGc1TnpzeE5qY3dOVFV4TURnNE9EazNPekU1T2tOSmFsVkRNV1pCUzJWeVQwVmlablk1ZWt0alVqZzBPV1YwVGt4amFEWTNPVkYzWldkbGEyTXRaMUV4UUhSb2NtVmhaQzUwWVdOMk1nPT18N2RjMTIzMWQwZDAyNDUwNzliNjMwOGRhZDk4OGQyODd8ZWYwNmJkMDY5NjNkNDk5ZGJlNWQxNWNlNWQyOGE0MmE%3D&sdata=V0ZXMG5qQUVnTUdORUpnQ2FibHVBNWtaeVdxVkE5Nk13a1dPUEcrY1RnTT0%3D).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,18 @@
+Contributing
+============
+
+Contributions are welcome from the community. Questions can be asked on the
+[issues page][1]. Before creating a new issue, please take a moment to search
+and make sure a similar issue does not already exist. If one does exist, you
+can comment (most simply even with just a `:+1:`) to show your support for that
+issue.
+
+If you have direct contributions you would like considered for incorporation
+into the project you can [fork this repository][2] and
+[submit a pull request][3] for review.
+
+
+
+[1]: https://github.com/USGS-R/drb-inland-salinity-ml/issues
+[2]: https://help.github.com/articles/fork-a-repo/
+[3]: https://help.github.com/articles/about-pull-requests/
--- a/code.json
+++ b/code.json
@@ -1,0 +1,75 @@
+[
+  {
+    "name": "drb-inland-salinity-ml",
+    "organization": "U.S. Geological Survey",
+    "description": "Code repo for Delaware River Basin machine learning models that predict inland salinity.",
+    "version": "main",
+    "status": "Development",
+    "permissions": {
+      "usageType": "openSource",
+      "licenses": [
+        {
+          "name": "Public Domain, CC0-1.0",
+          "URL": "https://github.com/USGS-R/drb-inland-salinity-ml/blob/main/LICENSE.md"
+        }
+      ]
+    },
+    "homepageURL": "https://github.com/USGS-R/drb-inland-salinity-ml/",
+    "downloadURL": "https://github.com/USGS-R/drb-inland-salinity-ml/archive/refs/heads/main.zip",
+    "disclaimerURL": "https://github.com/USGS-R/drb-inland-salinity-ml/blob/main/DISCLAIMER.md",
+    "repositoryURL": "https://github.com/USGS-R/drb-inland-salinity-ml.git",
+    "vcs": "git",
+    "laborHours": -1,
+    "tags": [
+      "R",
+      "USGS"
+    ],
+    "languages": [
+      "R",
+      "HTML"
+    ],
+    "contact": {
+      "name": "jds485",
+      "email": "FILL ME IN"
+    },
+    "date": {
+      "metadataLastUpdated": "2023-04-07"
+    }
+  },
+  {
+    "name": "",
+    "organization": "U.S. Geological Survey",
+    "description": "",
+    "version": "v0.0.0",
+    "status": "Ideation",
+    "permissions": {
+      "usageType": "openSource",
+      "licenses": [
+        {
+          "name": "Public Domain, CC0-1.0",
+          "URL": "LICENSE.md"
+        }
+      ]
+    },
+    "homepageURL": "",
+    "downloadURL": "",
+    "disclaimerURL": "",
+    "repositoryURL": "",
+    "vcs": "git",
+    "laborHours": 1,
+    "tags": [
+      "R",
+      "USGS"
+    ],
+    "languages": [
+      "R"
+    ],
+    "contact": {
+      "name": "",
+      "email": ""
+    },
+    "date": {
+      "metadataLastUpdated": "2022-10-16"
+    }
+  }
+]
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,9 @@
+Code of Conduct
+===============
+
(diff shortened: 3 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,18 @@
...
+can comment (most simply even with just a `:+1:`) to show your support for that
+issue.
+
+If you have direct contributions you would like considered for incorporation
+into the project you can [fork this repository][2] and
+[submit a pull request][3] for review.
+
+
+
+[1]: https://github.com/USGS-R/drb-inland-salinity-ml/issues
```
lekoenig, line 16: If we are transitioning this repo to GitLab and not DGEC, we'll have to remember to update these url's whenever that happens (here, and the url's in the code.json file).

code.json
```diff
@@ -0,0 +1,45 @@
...
+          "URL": "https://raw.githubusercontent.com/USGS-R/drb-inland-salinity-ml/main/LICENSE.md"
+        }
+      ]
+    },
+    "homepageURL": "https://github.com/USGS-R/drb-inland-salinity-ml/",
+    "downloadURL": "https://github.com/USGS-R/drb-inland-salinity-ml/archive/refs/heads/main.zip",
+    "disclaimerURL": "https://raw.githubusercontent.com/USGS-R/drb-inland-salinity-ml/main/DISCLAIMER.md",
+    "repositoryURL": "https://github.com/USGS-R/drb-inland-salinity-ml.git",
+    "vcs": "git",
+    "laborHours": -1,
```
lekoenig, line 22: I don't know how this metadata field is used, but should we add some ballpark here (or delete it)?

code.json
```diff
@@ -0,0 +1,45 @@
...
+    "tags": [
+      "R",
+      "USGS",
+      "usg-artificial-intelligence",
+      "research",
+      "random forest",
+      "explainable AI",
+      "specific conductivity",
+      "rivers"
```
lekoenig, line 31: I think the tags you have are great, so up to you whether to add the suggested text below. It looks like `usg-artificial-intelligence` tag is a known or pre-existing tag? If so, does adding that tag come with any additional guidance or compliance requirements for our repo?

```suggestion
      "salinity",
      "river"
```

lekoenig (approved):
Looks great! I've added a couple questions. Feel free to take a look at those, but I approve this PR as is.

code.json
jesse-ross: -1 is good. Don't delete it! Most folks leave it at the default. According to Eric Martinez, the main use case for actually using this would be if you had contracted with someone for a given number of hours. Since most of us don't have useful or accurate information to use to populate this, it uses -1 as an NA.

lekoenig: Thanks for the context, I promise I won't delete it! 😸

jds485: If we can come up with a good estimate, yes. I was not tracking programming hours, but we could probably get close with the hours we were collectively assigned to PUMP

code.json
lekoenig: OK, thanks! Good to know about that `usg-artificial-intelligence` tag. I missed that when looking over the guidance before.

Issue #233, opened 2023-04-12
Add monthly, seasonal error maps for each model


jds485:
Add monthly and seasonal residual error maps

Also, add reservoir locations to all spatial error maps

PR #238, opened 2023-04-14
Add maps of error metrics for different time periods

jds485:
This PR adds maps of RMSE by calendar month, season, and for each calendar year in the dataset. There is a new function that replaces the previous targets {commands} to make the code a little cleaner for these targets.

Currently, the plot does not have a label for what is plotted in it. That information is only within the file name. I think that's enough for our needs, just wanted to note it.

I also added reservoir locations to these maps as black circles. I had trouble getting those to display in the legend because the existing legend (RMSE) is for a continuous variable.

Example plot for year 2000 (one of the very high residual years):
RMSE_log10_year_2000_full

Closes: #233


lekoenig (commented):
Looks good, @jds485! I made one code suggestion to add the reservoirs item to the figure legend. 

Overall, I'm also wondering to what extent the functions `plot_nhdv2_attr` and `plot_maps` are expected to serve multiple purposes within this pipeline? For example, the `plot_nhdv2_attr` function now looks a lot like the synoptic viz we'd created previously and which lives on a separate branch in the repo. Given that, I wondered whether it'd be better to replace `"reservoirs"` with something more generic like `"sites"` and/or `"site_type"` so that separate functions aren't needed for those other applications. However, I don't have a good sense for what other site applications are needed _and_ recognize that generalizing this function may not be the best use of our limited time. So I'm OK with merging this PR as is once reservoirs are added to the figure legends. 

> Currently, the plot does not have a label for what is plotted in it. That information is only within the file name. I think that's enough for our needs, just wanted to note it.

One last question - is this something you were hoping to add but just not sure how to implement? If so, we could pass the info contained in `filename_end` to `ggtitle` as a variable.

jds485:
Thanks, Lauren! 

> Overall, I'm also wondering to what extent the functions `plot_nhdv2_attr` and `plot_maps` are expected to serve multiple purposes within this pipeline?

I think `plot_nhdv2_attr` is used for all maps and I've been thinking of this as generic. `plot_maps` is a wrapper around that function for different ways to aggregate information to plot on the map, and I also view that as generic. I forgot about the site vizualizations in the synoptic_site_viz branch. I'm also okay with keeping that separate, but it is a similar function. I don't intend on merging that branch with the main repo.

> Is this something you were hoping to add but just not sure how to implement?

No, I wasn't planning to add this information to the plots.

lekoenig (approved):

jds485:
I am going to wait to rebuild targets until merging #240 because of the merge conflict. I most recently ran #240 targets, and I think if I rebuild these targets I will need to rebuild in #240 as well

Issue #188, opened 2022-10-21
Add function to evaluate model performance for different attribute conditions


jds485:
Example: RMSE for high urban density vs. low urban density

Example targets with model predictions to try this with:
`p4_pred_RF_static_dynamic_temporal_test`
`p4_pred_RF_min_static_dynamic_temporal_test`
`p4_pred_RF_dynamic_temporal_test`

The attribute data for these targets are available in their model prediction target:
`p4_train_RF_dynamic_temporal$best_fit$splits[[1]]$data`
You'll have to join by `PRMS_segid` and `Date`

The function can compute the error metric according to splits in a provided attribute value (e.g., urban land cover < 0.25, 0.25-0.5, 0.5-0.75, and >=0 .75).
If it's simpler, could do the split and plotting together within, for example, the barplot function.

jds485:
@lekoenig: I think this would be the next most helpful to have for AGU

PR #240, opened 2023-04-14
Split error metrics by season, land cover, physio region

jds485:
This PR computes error metrics for each of the land cover, season, and physio data splits.

Closes: #188


Snakefile_metrics.smk
galengorski, line 1: What is the reason behind computing metrics within a snakemake framework instead of sticking with targets for this?

galengorski (approved):
@jds485 this looks good to me, I had a few clarifying questions which we discussed over teams chat. I'll paste the conversation below so that it's recorded here as well:

Galen:

- The function `write_preds_obs_splits` I'm not quite sure what the split parameter signifies here? 
- I'm also having a little trouble in pinning down the linkages between the snakemake workflow and the targets, the targets in 4_predict.R are writing the .txt and .csv files for results and metrics, respectively correct? Then the snakemake rules (e.g. `write_preds_obs_1`) what are these rules doing? 
- I'm also having trouble finding where the model_metrics function is being defined

Jared reply:
- `write_preds_obs_splits`: what does the split parameter signifies here?
- The split parameter is a dataframe of row indices (segment and date) that are used to select observations
- The `4_predict.R` targets provide txt and csv files that are used to compute metrics in the snakemake workflow for different data partitions (high forest, high urban, etc.).
- `model_metrics` is in `utils.py` and for me shows at the bottom of the "files changed" tab of the PR

Snakefile_metrics.smk
jds485: The primary reason was time to implement. The functions I used were available in river-dl and already being used for the RGCN model. I guess they could be called within targets using reticulate or a system call to snakemake instead. I've not tried that out but it would make it cleaner to keep within one pipeline

